PR #60, opened 2014-11-08
Add gfx-rs's cube example

tomaka:


tomaka:
We have a teapot example now (#261), which is better.

PR #83, opened 2014-11-24
Add possibility to rebuild the display with other options

tomaka:
Doesn't really work well for the moment


tomaka:
Still doesn't work for unknown reasons.

tomaka:
It's working :tada:

PR #194, opened 2015-01-01
Extract operations from fbo.rs into ops.rs

tomaka:


tomaka:
Build fails because of breaking changes in rustc. Tested locally, force pushing.

PR #196, opened 2015-01-02
Make cgmath and nalgebra optional features

tomaka:


tomaka:
Tested locally with an older rustc. Also `cargo test --no-default-features --features "headless"` passes. Force-pushing.

PR #198, opened 2015-01-02
Update for rustc

tomaka:


tomaka:
The linux nightlies haven't caught up yet.

PR #202, opened 2015-01-02
Remove travis filtering in tests/texture.rs

tomaka:


tomaka:
Failure:

> Mesa 8.0.4 implementation error: unexpected format GL_RGBA32F in _mesa_choose_tex_format()
> Please report at bugs.freedesktop.org

The bug is already fixed in newer versions of Mesa.

PR #209, opened 2015-01-03
Fix drawing with CPU indices sometimes using a stale pointer

tomaka:


tomaka:
Travis fails because of a breaking change in Rustc (again). Tested locally, force-pushing.

PR #212, opened 2015-01-04
Fixed a few of the rustc-change compile errors.

TyOverby:
There are still glium compile errors, but this covers a large swath of them.


tomaka:
Thanks

src/lib.rs
```diff
@@ -46,8 +46,8 @@ extern crate glium_macros;
 #[vertex_format]
 #[deriving(Copy)]
 struct Vertex {
-    position: [f32, ..2],
-    color: [f32, ..3],
+    position: [f32, 2],
+    color: [f32, 3],
```
tomaka: You forgot to replace `,` by `;`, but I'll fix that myself.

PR #214, opened 2015-01-04
Fix compilation with all features enabled

tomaka:


tomaka:
Let's pretend this passes.

PR #217, opened 2015-01-05
Improve tests stability on headless windows

tomaka:


tomaka:
Apparently this change doesn't improve anything.

PR #223, opened 2015-01-05
Rework how FBOs work internally

tomaka:
Extract everything related to FBOs in the `fbo.rs` module with a cleaner interface.

Will make reading textures/renderbuffers far easier.

### DIFF
--- a/src/fbo.rs
+++ b/src/fbo.rs
@@ -1,8 +1,8 @@
+use std::collections::HashMap;
 use std::mem;
-use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


tomaka:
Segfaults when used in a more complex project. Needs investigations.

tomaka:
Segfault seems specific to my windows machine.

tomaka:
Segfault was due to FBOs not being cleaned up. After around 375 objects the driver crashes. It's fixed now.

PR #238, opened 2015-01-11
Remove gl_extensions restriction for reading textures

tomaka:
Instead of calling `glGetTexture`, binds a framebuffer and use `glReadPixels`.

Introduces a potential, rare, race condition when destroying a texture that has been bound to the same FBO as the texture you are reading simultaneously. Will open an issue about it.


tomaka:
Opened #239

PR #247, opened 2015-01-12
Frame no longer borrows the Display

### LINKED ISSUE #215: Why does frame need to have the same lifetime as the Display that spawned it?
It makes a copy of the display to store inside itself, so it seems like the only purpose would be to keep you from performing operations on your original display.

tomaka:
Close #215


src/lib.rs
```diff
@@ -1001,7 +1001,6 @@ pub struct BlitHelper<'a>(&'a Arc<DisplayImpl>, Option<&'a fbo::FramebufferAttac
 /// instantaneous, even when vsync is enabled.
 pub struct Frame<'a> {
```
TyOverby: Shouldn't this <'a> be removed too?

PR #261, opened 2015-01-13
Add basic utah teapot example

tomaka:
The example was added mostly for the model. There's no texture, no movement, etc.


palodequeso:
Sweet! I didn't get a chance to work on that deferred example last night, but I'm going to work on it tonight.

tomaka:
I've added some lighting to the teapot. However it doesn't have normals or adjacency infos, so I had to use only one normal per face.

palodequeso:
Computing the normals isn't too hard. But I suppose that's out of the scope of Glium.

tomaka:
Yeah I could probably just modify `teapot_model` to add a pass.

palodequeso:
I mean, even if you generated them offline and changed the model. That way
you wouldn't need that geometry shader just to get some basic lighting.

tomaka:
That's what I was thinking. However I don't want to bother rewriting the Rust code corresponding to a new model.

Instead I'm going to add some code in `teapot_model` that builds the normals and returns them as part of the vertex.

palodequeso:
That makes sense.

PR #262, opened 2015-01-13
Update from original

drewet:


tomaka:
I don't think that was supposed to land here?

PR #263, opened 2015-01-13
Improve the lighting of the teapot

tomaka:
Follow-up of #261 

Lighting is slightly improved, but there may be something wrong with either my calculation or the model.

screenshot


examples/teapot.rs
```diff
@@ -40,46 +50,19 @@ fn main() {
         "
             #version 330
```
palodequeso: I think you can drop this to 130 now.

palodequeso:
I think when you are computing normals per face, and then averaging them per vertex, you are not guaranteeing that you are computing them with adjacent triangles, which I think may be the cause of some of the issues. Unless you are positive that the triangles are all adjacent when computing the final average normal across several faces, you'll have to find those adjacent triangles.

tomaka:
For the problem it's maybe because I don't have depth tests activated because windows doesn't give me a depth buffer.
I didn't think it was possible *not* to have a depth buffer. I guess I'll need to fix this in glutin.

Issue #258, opened 2015-01-12
[Help wanted!] Fix all the mistakes in the documentation


tomaka:
I'm opening an issue because there are a lot of grammatical mistakes in the documentation. Any help to fix these would be appreciated!

camjackson:
Still making progress on this, just got a few modules left.

camjackson:
This can probably be closed now.

PR #269, opened 2015-01-14
Fixed spelling and grammar in most files

camjackson:
More proof reading for #258. Still a few more files to go.


src/index_buffer.rs
```diff
@@ -2,7 +2,7 @@
 In order to draw, you need to provide a source of indices which is used to link the vertices
 together into *primitives*.
 
-There are height types of primitives, each one with a corresponding struct:
```
tomaka: I can't believe I wrote that :o

src/uniforms/sampler.rs
```diff
@@ -69,10 +69,10 @@ pub enum MinifySamplerFilter {
     /// Takes the nearest texel from the two nearest mipmaps, and merges them.
     LinearMipmapNearest,
 
-    /// Same as `Linear` but from the nearest mipmap.
+    /// Same as `Linear`, but from the nearest mipmap.
     NearestMipmapLinear,
 
-    ///
+    /// Same as `Linear`, but from the two nearest mipmaps.
```
camjackson: This was missing, I looked it up in the GL docs, so I'm pretty sure it's correct.

src/lib.rs
```diff
@@ -234,7 +234,7 @@ trait GlObject {
     fn get_id(&self) -> gl::types::GLuint;
 }
 
-/// Internal trait for enums that can be turned into GLenum.
+/// Internal trait for enums that can be turned into GLEnum.
```
tomaka: That should remain `GLenum`, as it's the name of the type (see three lines below).

tomaka:
Thanks a lot for doing that :heart_eyes:

src/index_buffer.rs
camjackson: Haha, guessing it was meant to say eight

src/lib.rs
camjackson: Ah OK, I changed it because of the trait with capital E, although that also has a lower-case L. I'll put it back :)

camjackson:
That's everything except for `tests`.

tomaka:
Note that you should rebase on master instead of merging.

Can I merge this or are you adding other things to this PR?

camjackson:
Ah right, sorry, didn't occur to me until after I'd merged it.

I'm all done now.

PR #275, opened 2015-01-14
Implement uniform buffers

tomaka:
Buffers that contain a single value (usually a struct) whose purpose is to be bound to a uniform block.
Not really useful for the moment as you can't bind them.

### DIFF
--- a/src/uniforms/buffer.rs
+++ b/src/uniforms/buffer.rs
@@ -1,0 +1,94 @@
+use Display;
+use buffer::{self, Buffer};
+
+use std::ops::{Deref, DerefMut};
+
+use context;
+use gl;
+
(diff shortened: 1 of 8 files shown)


src/uniforms/buffer.rs
```diff
@@ -0,0 +1,77 @@
...
+                                                             gl::STATIC_DRAW);
+
+        UniformBuffer {
+            buffer: buffer,
+        }
+    }
+
+    /// Uploads data in the uniforms buffer.
+    pub fn new_if_supported(display: &Display, data: T) -> Option<UniformBuffer<T>> {
+        None        // FIXME:
```
tomaka: Unfinished

PR #279, opened 2015-01-14
Implement persistent buffers mapping

tomaka:
Remains to do:
 - Tests
 - Add a fence when using the VB/IB/UB to draw


tomaka:
Everything is done, but `glMapBufferRange` returns null.

tomaka:
Fixed

PR #298, opened 2015-01-16
Rename the vertex_buffer module to just "vertex"

tomaka:
The `vertex_buffer` module contains more than just the vertex buffer.


palodequeso:
+1

PR #299, opened 2015-01-16
Add possibility to retreive a program's binary representation

tomaka:
- Adds a `gl_program_binary` Cargo feature.
- Adds a `Binary` struct in the `program` module.
- Adds `program.get_binary()` and `program.get_binary_if_supported()`.
- Doesn't allow creating a program from the binary yet.


palodequeso:
+1

Issue #301, opened 2015-01-17
Need to support glBlendFunc(GL_ONE, GL_ONE); for deferred rendering example.


palodequeso:
In order to do the additive blending when iterating over the scene's lights and rendering to the light FBO, I need to be able to specify a blending function of glBlendFunc(GL_ONE, GL_ONE); Currently that is not an option.

tomaka:
Fixed by #303

palodequeso:
Thanks, I should be able to finish this example up today now.

Issue #161, opened 2014-12-26
Implement all the blending operations


tomaka:

bvssvni:
Could add a struct `Blend`:

```Rust
pub struct Blend {
    pub color: BlendingFunction,
    pub alpha: BlendingFunction,
    pub color: (f32, f32, f32, f32),
}

impl Default for Blend {
     fn default() -> Self {
        Blend {
            color: BlendingFunction::Addition {
                source: LinearBlendingFactor::SourceAlpha,
                destination: LinearBlendingFactor::OneMinusSourceAlpha,
            },
            alpha: BlendingFunction::Addition {
                source: LinearBlendingFactor::One,
                destination: LinearBlendingFactor::One,
            },
            color: (1.0, 1.0, 1.0, 1.0),
        }
    }
}
```

This will make alpha blending become enabled by default if `DrawParameters::blend` is set to `Some(Default::default())`.

Then replace `DrawParameters::blending_function` with

```Rust
pub blend: Option<Blend>
```

bvssvni:
Add the following to `LinearBlendingFactor`:

```Rust
SourceAlphaSaturated,
ConstColor,
ConstAlpha
OneMinusConstColor,
OneMinusConstAlpha,
```

tomaka:
I don't think `Default` should be implemented for the blending function.
Sure the regular alpha blending is what most people will do, but it's also too magical.

Having a `Blend` struct is not a bad idea though.

bvssvni:
Ok, so no default?

tomaka:
Yeah, no `Default`.

I think that a `Blend::alpha_blending()` method would be nice though.

bvssvni:
I'll make a PR.

bvssvni:
- [ ] Add struct `Blend`
- [ ] Derive `Clone, Debug, PartialEq` for `Blend`
- [ ] Move over doc comment from `BlendFunction` to `Blend` (and fix to reflect the new API)
- [ ] Add `Blend::alpha_blending`

bvssvni:
I've talked to @tomaka about https://github.com/tomaka/glium/issues/1208 and we decided to remove the `Option` to get rid of the "don't care" state, which might lead to confusion.

PR #303, opened 2015-01-17
Rework blending to make it more flexible

### LINKED ISSUE #302: Implement per-buffer blending functions
In OpenGL, you can specify a different blending function for each target you are rendering to.

tomaka:
Close #302 
cc #161 

Breaking change: removes `LerpBySourceAlpha`. Replace by `Addition { source: Alpha, destination: OneMinusSourceAlpha }`.


tomaka:
Oops, meant to close #301 instead of #302

PR #306, opened 2015-01-17
Adding a deferred shading example to Glium

palodequeso:
@tomaka 

Adding a very simple deferred pipeline example to glium. It only does diffuse lighting with attenuation for now.

### DIFF
--- a/examples/deferred.rs
+++ b/examples/deferred.rs
@@ -1,0 +1,379 @@
+#![feature(plugin)]
+
+#[plugin]
+extern crate glium_macros;
+extern crate glutin;
+extern crate glium;
+#[cfg(feature = "cgmath")]
+extern crate cgmath;
+#[cfg(feature = "image")]
+extern crate image;
+
+use glium::Surface;
+use glium::DisplayBuild;
+#[cfg(feature = "cgmath")]
+use cgmath::FixedArray;
+use std::io::BufReader;
+
+#[cfg(not(all(feature = "cgmath", feature = "image")))]
+fn main() {
+    println!("This example requires the `cgmath` and `image` features to be enabled");
+}
+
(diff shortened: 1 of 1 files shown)


tomaka:
Yay \o/

The travis failure is due to a change in Cargo. Could you rebase over master to make it pass?

palodequeso:
Sure thing, as soon as my daughter lays down for her nap. :)

palodequeso:
> Also, there's no copyright on this texture?

That's a good question, it didn't even cross my mind. I'll just switch to using the one you provided.

tomaka:
Oh and it doesn't work because the `cgmath` dependency is optional.

Unfortunately there's no way to run the tests without compiling the examples, so you need to do like I did here: https://github.com/tomaka/glium/blob/master/examples/image.rs#L18
With `#[cfg(all(feature = "cgmath", feature = "image"))]`

palodequeso:
Got it, I made the change and am running the tests locally first this time. :p

palodequeso:
When I try to run them locally, I get an error saying I have a duplicate definition of value main. I'm guessing I need to enable those features somehow before running cargo test?

tomaka:
The tests should pass with both `cargo test` and `cargo test --no-default-features`.

Your file should look like this:
```rust
#[cfg(not(all(feature = "cgmath", feature = "image")))]
fn main() {
    println!("This example requires the `cgmath` and `image` features to be enabled");
}

#[cfg(all(feature = "cgmath", feature = "image"))]
fn main() {
    ...  the real code here ...
}
```

palodequeso:
I made the changes and squashed my changes down to one commit.

palodequeso:
Yay, all is well!

PR #331, opened 2015-01-20
Reorganize the order of the shaders list

tomaka:
Put the shaders list in the order in which they are executed in the pipeline.

Doesn't change anything in the API.


palodequeso:
I thought geometry was the stage before tessellation?

palodequeso:
Never mind, I am wrong.
+1

PR #340, opened 2015-01-21
Implement instancing

tomaka:
- Adds an `gl_instancing` Cargo feature.
- Adds `PerInstanceAttributesBuffer` (and `PerInstanceAttributesBufferAny`) which can be used as a vertices source.
- If one of them is passed, the number of instances corresponds to the number of elements in the `PerInstanceAttributesBuffer`.
- Adds a variant in `DrawError` if passing multiple `PerInstanceAttributesBuffer`s and they don't have the same number of elements.

Example:

```rust
let buffer1 = {
    #[vertex_format]
    #[derive(Copy)]
    struct Vertex {
        position: [f32; 2],
    }

    glium::VertexBuffer::new(&display, 
        vec![
            Vertex { position: [-1.0,  1.0] },
            Vertex { position: [ 1.0,  1.0] },
            Vertex { position: [-1.0, -1.0] },
            Vertex { position: [ 1.0, -1.0] },
        ]
    )
};

let buffer2 = {
    #[vertex_format]
    #[derive(Copy)]
    struct Vertex {
        color: [f32; 3],
    }

    glium::vertex::PerInstanceAttributesBuffer::new(&display, 
        vec![
            Vertex { color: [0.0, 0.0, 1.0] },
            Vertex { color: [0.0, 1.0, 0.0] },
            Vertex { color: [1.0, 0.0, 0.0] },
        ]
    )
};

target.as_surface().draw((&buffer1, &buffer2), &index_buffer, &program, &uniform!{},
                          &std::default::Default::default()).unwrap();

// draws 3 instances (one per element in `buffer2`) of 4 vertices
//   each (one per element in `buffer1`)
// the `color` attribute in the vertex shader matches the elements in buffer2
```

~~Not working for now (despite what travis says), need to investigate.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,12 @@ script:
     - cargo build --verbose
     - cargo test --features "headless" --verbose
     - cargo test --no-default-features --features "headless" --verbose
-    - cargo build --features "headless gl_read_buffer gl_uniform_blocks gl_sync gl_persistent_mapping gl_program_binary gl_tessellation" --verbose
+    - cargo build --features "headless gl_read_buffer gl_uniform_blocks gl_sync gl_persistent_mapping gl_program_binary gl_tessellation gl_instancing" --verbose
 
 after_success: |
     [ $TRAVIS_BRANCH = master ] &&
     [ $TRAVIS_PULL_REQUEST = false ] &&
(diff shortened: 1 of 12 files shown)


tomaka:
Need to:

[ ] Write an example (with for example 10k cubes)
[ ] Add an example on the website

tomaka:
The example doesn't behave exactly as expected.

PR #342, opened 2015-01-21
Update for UFCS

tomaka:
Removes these dirty hacks with `Option` now that they are no longer needed.


tomaka:
Conflicts with #340

PR #344, opened 2015-01-21
Implement clear() in MultiOutputFrameBuffer

tomaka:


palodequeso:
The build says it failed but it doesn't have any console output, weird.

palodequeso:
+1 once the build passes.

Issue #343, opened 2015-01-21
Support OpenGL version < 2


tomaka:
With the `GL_ARB_vertex_shader`, `GL_ARB_fragment_shader`, `GL_ARB_vertex_buffer_object`, `GL_ARB_framebuffer` extensions, it should be possible to support any version of OpenGL.

However more things may be taken into account, like `GL_ARB_non_power_of_two`.

tomaka:
The restriction is still there because textures don't work.

This is probably related to texture formats.

katyo:
I trying to create grayscale (one component) texture (OpenGL ES 3.0) and get INVALID_OPERATION:

PR #349, opened 2015-01-22
Use associated types in GlObject trait

tomaka:
Will be needed because the `GL_ARB_shader_objects` extension uses `GLhandleARB` and not `GLuint`.

cc #343


tomaka:
Annoying travis errors :-/

PR #350, opened 2015-01-22
Support GL_ARB_shader_objects

tomaka:
cc #343

Remains to do:
- Fix the bugs (apparently I broke something)
- Use `glUniform*ARB` instead of `glUniform`

### DIFF
--- a/src/program.rs
+++ b/src/program.rs
@@ -5,8 +5,9 @@ use std::collections::HashMap;
 use std::sync::{Arc, Mutex, StaticMutex, MUTEX_INIT};
 use std::sync::mpsc::channel;
 use {Display, DisplayImpl, GlObject};
-use context::{self, CommandContext};
+use context::{self, CommandContext, GlVersion};
 use uniforms::UniformType;
+use Handle;
 
 /// Some shader compilers have race-condition issues, so we lock this mutex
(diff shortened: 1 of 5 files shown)


tomaka:
With this PR, the triangle example works on a machine that only has OpenGL 1.4 available.

src/program.rs
```diff
@@ -650,13 +859,27 @@ unsafe fn reflect_uniforms(ctxt: &mut CommandContext, program: gl::types::GLuint
...
 
-    let mut active_attributes: gl::types::GLint = mem::uninitialized();
-    ctxt.gl.GetProgramiv(program, gl::ACTIVE_ATTRIBUTES, &mut active_attributes);
+    // number of active attributes
+    let active_attributes = {
+        let mut active_attributes: gl::types::GLint = mem::uninitialized();
+        match program {
+            Handle::Id(program) => {
+                assert!(ctxt.version >= &GlVersion(2, 0));
+                ctxt.gl.GetProgramiv(program, gl::ACTIVE_UNIFORMS, &mut active_attributes);
```
tomaka: `gl::ACTIVE_UNIFORMS` should be `ACTIVE_ATTRIBUTES`

PR #351, opened 2015-01-22
Make the event iterator actually infinite

aepsil0n:
I had a problem with the following use case:

```rust
let display = display.clone();
Thread::spawn(move || {
    for event in display.poll_events() {
        println!("receive {:?}", event);
    }
    println!("die");
});
```

The output was a single resize on start-up and then it would "die", even though the application was still running.

This change fixes it for me, but I am unsure whether this touches any other multi-threading issues.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -41,12 +41,12 @@ pub struct CommandContext<'a, 'b> {
 }
 
 /// Iterator for all the events received by the window.
-pub struct EventsIter<'a> {
+pub struct PollEventsIter<'a> {
     context: &'a Context,
     current_rx: Option<Receiver<glutin::Event>>,
 }
 
-impl<'a> Iterator for EventsIter<'a> {
+impl<'a> Iterator for PollEventsIter<'a> {
     type Item = glutin::Event;
 
     fn next(&mut self) -> Option<glutin::Event> {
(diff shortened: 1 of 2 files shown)


src/context/mod.rs
```diff
@@ -379,19 +381,28 @@ impl Context {
...
-
-                            // sending back
-                            notify.send(event).ok();
-                        }
+                        notifiers.push(notify)
                     },
                     Err(_) => break
                 }
+
+                for event in window.poll_events() {
```
tomaka: This code is only executed when the window is destroyed. I don't see how that fixes it :huh:

tomaka:
What you need is `wait_events` instead, no?
I should have read what you needed more carefully.

src/context/mod.rs
aepsil0n: Are you sure? But I am continuously receiving all the events… I thought, this is run every frame.

aepsil0n:
Oh, I see. Sorry, I thought you had intended to change `poll_events` to be infinite. Ok, that makes more sense. Let me update that, so that there is both a `wait_events` and a `poll_events`.

src/context/mod.rs
tomaka: I changed the architecture in a previous change.

tomaka:
You probably just need to duplicate the `EventsIter` struct, no need to change anything in the internal loop.

tomaka:
Having a spinlock is bad, but I guess there's no other solution for now :-/

Thanks!

aepsil0n:
Yes, this is why started modifying the internal loop. But you wanted to redesign the whole thing anyway, right? ;)

PR #379, opened 2015-01-26
Do not use glTexStorage*D with compressed textures

tomaka:


palodequeso:
that looks better

tomaka:
Closing in favor of #380

PR #380, opened 2015-01-27
Texture formats, take 2

tomaka:


tomaka:
cc @palodequeso 
This may have solved your problem

palodequeso:
It looks like it fixed everything but a depth format error:
```
Debug message: `FS SIMD8 shader: 9 inst, 0 loops, compacted 144 to 96 bytes.
`
Debug message: `FS SIMD16 shader: 9 inst, 0 loops, compacted 144 to 96 bytes.
`
Debug message: `VS vec4 shader: 8 inst, 0 loops, compacted 128 to 128 bytes.
`
thread '<main>' panicked at 'Only ClientFormat::F32 can be used to upload on a depth texture', /home/douglas/.cargo/git/checkouts/glium-cee1c79709e4d1bd/master/src/texture/tex_impl.rs:499
```

tomaka:
Should be fixed by #384

PR #384, opened 2015-01-27
Remove unnecessary constraints on new_empty

tomaka:
When creating an empty texture, you have to call `glTexImage2D(.., format, NULL);`

For the moment `new_empty` calls `TextureImplementation::new` with a dummy `U8U8U8U8` format. However with the recent checks that has been added, this sometimes fails for example with `Only ClientFormat::F32 can be used for depth textures` when you try to create an empty depth texture.

This fixes it.


palodequeso:
Sweet!

PR #385, opened 2015-01-27
Only generate mipmaps for color textures

tomaka:


palodequeso:
Ahh, now this makes a lot of sense. I'll give it a go as soon as it's ready.

tomaka:
Also stops generating mipmaps for compressed textures, should fix them.

palodequeso:
Hooray, all is well!

PR #389, opened 2015-01-27
Allow drawing without indices

tomaka:
- Adds `glium::index_buffer::NoIndices(PrimitiveType)` to use the vertices in the same order as they are in the vertices source.
- Removes some member functions of `IndicesSource`.
- Adds `DrawError::VerticesSourcesLengthMismatch` that is returned if you use `NoIndices` and multiple vertices sources with different sizes.


palodequeso:
+1

PR #394, opened 2015-01-28
Try adding appveyor continuous integration

tomaka:


tomaka:
Failed because of breaking changes in rustc.

tomaka:
Apparently the default 1.1 software implementation on windows doesn't support enough stuff. Closing.

PR #395, opened 2015-01-28
Pass --nocapture when running tests

tomaka:


tomaka:
The initial idea was to show debug messages produced by the driver.
However there are no such message, and all that is displayed are confusing panick messages for tests are have the `#[should_fail]` attribute.

I guess I'll close this.

PR #408, opened 2015-01-29
Add test for compressed textures drawing

tomaka:


tomaka:
That may finally close this compressed textures fiasco.

PR #418, opened 2015-01-30
Allow rewriting some parts of Texture2ds and CompressedTexture2ds

### LINKED ISSUE #417: Add support for blitting from CPU memory
It would be really useful to have support for copying CPU-side data into textures every frame. This would be useful for things like embedded web browsing via the Chromium Embedded Framework.

tomaka:
Adds `Texture2d::write` and `CompressedTexture2d::write`.

Close #417

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -435,6 +435,42 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
             "#)).unwrap();
     }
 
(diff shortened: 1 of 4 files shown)


dylanede:
Thank you for being so fast!
However, as you said it's probably faster to just generate a new texture every frame instead of editing existing ones.

tomaka:
I'll still merge this. Being able to write over an existing texture is useful too.

PR #429, opened 2015-02-02
Fix missing clear in the deferred example

### LINKED ISSUE #398: Deferred Example needs Clear before rendering prepass
Rather than forking and making a pull request, I'll just link to the fix in my own rendering code since it should only be a one liner: https://github.com/palodequeso/damage/blob/master/src/graphics/renderer.rs#L126

This way if anyone goes to use the deferred example for anything, and they start adding camera movement, they won't get weird smearing in the g-buffer.

tomaka:
Close #398


palodequeso:
sweet. +1

PR #440, opened 2015-02-03
Change texture-related traits

### LINKED ISSUE #388: Rework Texture*dData traits
New design proposal:

```rust
pub trait Texture2dData {
    /// Represents raw pixel data.
    type Data: Copy + Send = u8;

    fn as_raw<'a>(&'a self) -> RawTexture2dData<'a, Self::Data>;

    /// Returns the list of format preferred for `from_raw`.
    fn get_preferred_formats() -> Vec<ClientFormat>;

    fn from_raw(RawTexture2dData<Self::Data>) -> Self;
}

pub struct RawTexture2dData<'a, T> {
    /// The length of `data` must be equal to width * height * format.get_size() / mem::size_of::<T>().
    pub data: Cow<'a, Vec<T>, [T]>,
    pub width: u32,
    pub height: u32,
    pub format: ClientFormat,
}
```

Same for `Texture1dData` and `Texture3dData`.

tomaka:
Adds some `unimplemented!()` for texture arrays, but texture arrays aren't working anyway.

I also changed `get_preferred_formats()` to return a `Vec`, because `&'static [ClientFormat]` is too hard to get right. This will need some refinements.

Close #388

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -222,9 +222,9 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
     // writing the `new` function
     {
         let data_type = match dimensions {
-            TextureDimensions::Texture1d | TextureDimensions::Texture1dArray => "Texture1dData",
-            TextureDimensions::Texture2d | TextureDimensions::Texture2dArray => "Texture2dData",
-            TextureDimensions::Texture3d => "Texture3dData",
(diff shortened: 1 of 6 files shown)


dylanede:
How does this relate to #388? I am really considering attempting to write a patch for that, as the performance is quite bad in CEF with the current implementation. I also want to add support for updating textures with subregions (using GL_UNPACK_SKIP_PIXELS and GL_UNPACK_SKIP_ROWS) so that dirty rectangles from CEF can update efficiently. Is that alright?

PR #444, opened 2015-02-03
Reworked texture API to fix the copy-on-write part of #388.

dylanede:
**Edit: Please do not merge this until I have run further tests.**
This implements the desired texture API from #388 that cuts down on copies. In order to do so I have created an additional blocking version of `exec` called `exec_sync` that ensures that the context thread can safely use any references from the calling thread. Due to limitations with the current version of `Send` in Rust this has been implemented using `unsafe` code, but in this case the safety of the code is easily verified.

All tests pass, the examples work and the docs build.

Also note that my emacs has removed trailing whitespace in a few places. I can split this into a separate commit if necessary.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -25,8 +25,9 @@ enum Message {
 }
 
 pub struct Context {
-    commands: Mutex<Sender<Message>>,
-    sync: Mutex<SyncSender<()>>,
+    senders: Mutex<(
+        Sender<Message>,
+        SyncSender<Box<for<'a, 'b> ::std::thunk::Invoke<CommandContext<'a, 'b>, ()> + Send>>)>,
 
     window: Option<Arc<glutin::Window>>,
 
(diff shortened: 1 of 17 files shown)


src/context/mod.rs
```diff
@@ -90,6 +92,15 @@ impl Context {
...
         self.commands.lock().unwrap().send(Message::Execute(Box::new(f))).unwrap();
     }
+    pub fn exec_sync<'a, F>(&self, f: F) where F: FnOnce(CommandContext) + 'a {
+        use std::thunk::Invoke;
+        use std::mem::transmute;
+        let f: Box<for<'b,'c> Invoke<CommandContext<'b, 'c>>> = Box::new(f);
+        let pretend_sync_f: Box<for<'b,'c> Invoke<CommandContext<'b, 'c>> + Send> = unsafe{ transmute(f) };
+        self.commands.lock().unwrap().send(Message::Execute(pretend_sync_f)).unwrap();
+        self.commands.lock().unwrap().send(Message::Sync).unwrap();
+        self.sync.lock().unwrap().send(());
```
dylanede: This could be optimised to use only one mutex at a later date.

dylanede:
Just realised that this is not actually completely safe in the case that multiple threads are using the context. I will just write a patch for that.

dylanede:
At some point the project should be updated to the latest nightlies to please Travis.

PR #447, opened 2015-02-04
Fixed #388 and a thread desynchronisation issue in Context

### LINKED ISSUE #388: Rework Texture*dData traits

dylanede:
Reworked texture API to fix the copy-on-write part of #388 and fixed an issue where the main loop could run ahead of rendering due to not blocking on buffer swaps. This was possibly the hidden issue having the most effect on the performance of my CEF experiments. The CPU was running away queueing up commands for the GL thread without blocking, resulting in extreme lag between the display and user input. With this fixed and my other improvements performance using CEF with Glium is nearly as good as native Chrome.

The main change affecting the API is that everywhere a user would pass a data source they now pass a reference to one. Other than slightly affecting when memory is copied and deallocated, this shouldn't affect performance at all for existing use cases, but it allows direct copying from memory into mapped textures, useful for streaming data. The updated data source structs are nearly identical to your proposal in #388.

Internally, an additional method has been added to Context - exec_sync. This function can execute an arbitrary function on the GL thread and return synchronously. It also provides the benefits that the function does not need to be `Send`, and it can return a result (though that result has to be `Send`). This is used in texture creation and writing, but it is designed with the idea that it may be used for future code that needs lifetimes smaller than `'static` or return a result from the GL.

*Also, if you know how I can merge these commits together (rebase seems to ignore merge commits), please let me know*

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -239,7 +239,7 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
                 /// Builds a new texture by uploading data.
                 ///
                 /// This function will automatically generate all mipmaps of the texture.
-                pub fn new<T>(display: &::Display, data: {param})
+                pub fn new<T>(display: &::Display, data: &{param})
                               -> {name} where T: {data_type}
                 {{
             ", data_type = data_type, param = param, name = name)).unwrap();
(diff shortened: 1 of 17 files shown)


tomaka:
I thought you were fixing the row skip thingy, not the avoid-a-copy thing.

In my opinion, it's just wrong to force a synchronization when it isn't needed. What I thought I would do is simply implement the `Texture2dSource` on references, instead of requiring the user to pass a reference.

dylanede:
OK, I'll see what I can do to make the synchronisation only happen when needed.

PR #449, opened 2015-02-04
Modified data source traits to support taking Copy-on-Write data.

dylanede:
In the process, added optionally synchronising exec function to Context. Made texture creation and uploading synchronise with the GL thread if using shared data.

This patch should not affect existing user code that does not manually implement data sources.

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -239,8 +239,8 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
                 /// Builds a new texture by uploading data.
                 ///
                 /// This function will automatically generate all mipmaps of the texture.
-                pub fn new<T>(display: &::Display, data: {param})
-                              -> {name} where T: {data_type}
+                pub fn new<'a, T>(display: &::Display, data: {param})
+                              -> {name} where T: {data_type}<'a>
                 {{
             ", data_type = data_type, param = param, name = name)).unwrap();
 
@@ -287,7 +287,7 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
 
             TextureDimensions::Texture1dArray => (write!(dest, "
                     let array_size = 0;
-                    let data = Vec::<u8>::new();
+                    let data = ::std::borrow::Cow::Owned(Vec::<u8>::new());
                     let width = 0;
                     let client_format = unsafe {{ ::std::mem::uninitialized() }};
                     unimplemented!();
@@ -295,7 +295,7 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
 
             TextureDimensions::Texture2dArray => (write!(dest, "
                     let array_size = 0;
-                    let data = Vec::<u8>::new();
+                    let data = ::std::borrow::Cow::Owned(Vec::<u8>::new());
                     let width = 0;
                     let height = 0;
                     let client_format = unsafe {{ ::std::mem::uninitialized() }};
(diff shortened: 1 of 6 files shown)


build/textures.rs
```diff
@@ -287,7 +287,7 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
 
             TextureDimensions::Texture1dArray => (write!(dest, "
                     let array_size = 0;
-                    let data = Vec::<u8>::new();
+                    let data = ::std::borrow::Cow::Owned(Vec::<u8>::new());
```
tomaka: You don't need `::std::borrow::` as you're importing `Cow` in `texture/mod.rs`

build/textures.rs
```diff
@@ -295,7 +295,7 @@ fn build_texture<W: Writer>(mut dest: &mut W, ty: TextureType, dimensions: Textu
 
             TextureDimensions::Texture2dArray => (write!(dest, "
                     let array_size = 0;
-                    let data = Vec::<u8>::new();
+                    let data = ::std::borrow::Cow::Owned(Vec::<u8>::new());
```
tomaka: Same

src/texture/tex_impl.rs
```diff
@@ -47,8 +48,8 @@ pub struct TextureImplementation {
 
 impl TextureImplementation {
     /// Builds a new texture.
-    pub fn new<P>(display: &Display, format: TextureFormatRequest,
-                  data: Option<(ClientFormat, Vec<P>)>, width: u32, height: Option<u32>,
+    pub fn new<'a, P: Clone + 'a>(display: &Display, format: TextureFormatRequest,
```
tomaka: Could you add ` + Clone + 'a` in the `where` clause at the end of the function, since there's already one?

And while you're at it, realign the parameters vertically.

src/texture/tex_impl.rs
```diff
@@ -271,8 +273,8 @@ impl TextureImplementation {
     }
 
     /// Changes some parts of the texture.
-    pub fn upload<P>(&self, x_offset: u32, y_offset: u32, z_offset: u32,
-                     (format, data): (ClientFormat, Vec<P>), width: u32, height: Option<u32>,
+    pub fn upload<'a, P: Clone + 'a>(&self, x_offset: u32, y_offset: u32, z_offset: u32,
```
tomaka: Same as above

src/texture/tex_impl.rs
```diff
@@ -106,7 +107,8 @@ impl TextureImplementation {
         };
 
         let (tx, rx) = channel();
-        display.context.context.exec(move |: ctxt| {
+        //This always syncs with rx later.
+        display.context.context.exec_maybe_sync(false, move |: ctxt| {
```
tomaka: Hmm, what's the point of using `exec_maybe_sync` if you're always passing `false`?

dylanede:
Ok, I'll just fix that

tomaka:
Two or three minor modifications, but it's good otherwise :) Thanks!

dylanede:
I don't always pass false. Look at upload

tomaka:
I understand the point of `exec_maybe_sync`, but in the line where I added the comment, you explicitely wrote `display.context.context.exec_maybe_sync(false, move |: ctxt| {`.

tomaka:
I'll need to wait for travis to pass however ; it has already happened that travis detects bugs even when tests were passing locally. This won't happen until nalgebra compiles again with the nightly.

tomaka:
> I understand the point of exec_maybe_sync, but in the line where I added the comment, you explicitely wrote exec_maybe_sync(false, move |: ctxt| {.

Oh I got it, it's because `Send` is not implemented on the type.

dylanede:
I've just done a git amend and force push to remove the unnecessary namespacing, but github is refusing to show the changes. Could you possibly check out a local copy to see if the changes are actually there?

dylanede:
Also, if you're planning to get Travis to pass, it's probably a good idea to update to the latest nightly yourself and add feature attributes to stop the huge wall of warnings.

dylanede:
OK, it seems like the remote has updated, but the site for some reason is not showing the changes.

dylanede:
That should do it

tomaka:
You will also need to rebase over master to fix the compilation.

dylanede:
Only the build for nalgebra failed.

Issue #471, opened 2015-02-08
Texture API rework


tomaka:
Texture APIs are very clumbsy for the moment. Here is what I suggest in term of functionnalities.

```rust
impl TextureSomething {
    // returns the number of mipmaps, minimum 1
    fn get_mipmap_levels(&self) -> u32;

    // with `new`, glium chooses which format and whether you have mipmaps
    // non-array textures
    fn new<I>(data: I) -> Self where I: Texture*dData;
    // array textures
    fn new<I>(data: Vec<I>) -> Self where I: Texture*dData;

    // builds the texture with a specific format
    // `mipmaps` set to true does not necessarly mean you'll have mipmaps, but when it's
    // set to false you are sure not to get any
    // non-array textures
    fn with_format<I>(data: I, format: RelevantFormatType, mipmaps: bool)
                      -> Result<Self, TextureCreationError> where I: Texture*dData;
    // array textures
    fn with_format<I>(data: Vec<I>, format: RelevantFormatType, mipmaps: bool)
                      -> Result<Self, TextureCreationError> where I: Texture*dData;

    // builds the texture and specify whether you want mipmaps or not
    // `mipmaps` set to true does not necessarly mean you'll have mipmaps, but when it's
    // set to false you are sure not to get any
    // non-array textures
    fn with_mipmaps<I>(data: I, mipmaps: bool)
                       -> Self where I: Texture*dData;
    // array textures
    fn with_mipmaps<I>(data: Vec<I>, mipmaps: bool)
                       -> Self where I: Texture*dData;

    // all textures
    // tries to get the "best" format, with 4 components
    // replaces `new_empty`
    fn empty(dimensions...: RelevantDimensions...)
             -> Self;

    // all textures
    fn empty_with_format(format: RelevantFormatType, dimensions...: RelevantDimensions...)
                  -> Result<Self, TextureCreationError>;

    // non-array and non-cube
    // equivalent to `main_level().read()`
    fn read<I>(&self) -> I where I: Texture*dData;
    // array and cube textures
    fn read<I>(&self) -> Vec<I> where I: Texture*dData;

    // the `write` functions regenerate all mipmaps when it is called
    // non-array and non-cube textures only
    fn write<I>(&mut self, Rect, data: I) -> I where I: Texture*dData;

    // non-array and non-cube textures only
    // depending on the texture, the return type is different
    // call be called with level `0`
    fn mipmap<'a>(&'a self, level: u32) -> Option<MipmapSomething<'a>>;

[...]

PR #487, opened 2015-02-12
Add the empty, empty_with_mipmaps and empty_with_format functions

### LINKED ISSUE #486: Need api to disable mipmapping for textures attached to FBOs
Currently when creating an FBO texture, I use new_empty, but with this method, there is no wait to disable mip-mapping, which is something a user would definitely want control over when creating FBO attachments.

tomaka:
Close #486 
cc #471 

The `empty` method is like the current `new_empty` but doesn't need to pass any format and does not generate mipmaps.

The `new_empty` method has been deprecated.


palodequeso:
How does it know which format to use?

PR #489, opened 2015-02-12
Fix texture creation and add tests

tomaka:


tomaka:
This will need more work in order to know *exactly* which version supports which type of texture.

tomaka:
I have commented-out some tests. Will open related issues.

PR #493, opened 2015-02-13
Improve the way commands are being processed to remove allocations

tomaka:
Avoids all allocations when submitting a command by having a ring buffer with fixed-sized elements and panicking if the elements are too large.

### DIFF
--- a/src/context/commands.rs
+++ b/src/context/commands.rs
@@ -1,0 +1,186 @@
+use std::collections::RingBuf;
+use std::sync::{atomic, Arc, Condvar, Mutex};
+use std::{mem, ptr};
+use context::CommandContext;
+
+const CLOSURES_MAX_SIZE: usize = 64;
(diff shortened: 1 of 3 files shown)


tomaka:
I can't see any performance improvement with this change.
Profiling shows a ~30% decrease in the time spent in `je_mallocx` (which is normal since we are trying to avoid allocations), however the FPS count is approximately the same (+-2%) and debuggers don't seem to see any difference.

tomaka:
Will benchmark that again when all other allocations are removed when drawing as well.

tomaka:
I'm going to merge this as it fixes the commands buffer overflowing.

tomaka:
@nilbold This has fixed your problem

nilbold:
thanks much, everything seems to be running smoothly with or without a forced sleep call.

PR #530, opened 2015-02-19
[Breaking change] Add depth_write and rename DepthFunction to DepthTest

### LINKED ISSUE #433: Need a draw parameter for enabling glDepthMask.
This is pretty straight forward, if the parameter is true, then just add the state glDepthMask(GL_TRUE); to the OpenGL state stack. This allows one to do things like, render a scene in a deferred manner, but still render translucent or non-lit objects in a forward manner and combine the results without any extra shader magic.

tomaka:
- Renames `DepthFunction` to `DepthTest`
- Renames `depth_function` to `depth_test`
- Adds `depth_write`

Before this change, when a depth test passed the depth value of the pixel was always written on the depth buffer. This change adds a `depth_write` member to `DrawParameters` that allows you to set whether or not you want this.
**Its default value is `false`**. If you were using depth tests, you most likely should set `depth_write` to `true` in the parameters. See #433 for more details.

Close #433

### DIFF
--- a/src/context/state.rs
+++ b/src/context/state.rs
@@ -93,6 +93,9 @@ pub struct GLState {
     /// The latest value passed to `glDepthFunc`.
     pub depth_func: gl::types::GLenum,
 
+    /// The latest value passed to `glDepthMask`.
+    pub depth_mask: bool,
+
     /// The latest values passed to `glDepthRange`.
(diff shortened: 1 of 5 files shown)


palodequeso:
Works for me, looks like the build failed though.

PR #533, opened 2015-02-19
Fix syntax extension signatures

aepsil0n:


tomaka:
> The command "cargo test -j 1 --no-default-features --features "headless" --verbose" exited with 0.

Yay! Thanks.

Issue #34, opened 2014-11-03
Implement transform feedback


tomaka:

tomaka:
Some ideas:

- Add a `TransformFeedbackSession` struct which contains one or several `&mut VertexBuffer`.
- Add `TransformFeedbackSession::end(self) -> TransformFeedbackResult`.
- Add `VertexBuffer::to_transform_feedback_session()` which returns a `TransformFeedbackSession`.

- Add a `draw_primitives` member to `DrawParameters` of type `bool` (will probably be implemented in a separate PR).
- Add a `primitives_output` member to `DrawParameters` of type `Option<&TransformFeedbackSession>`.

- Add `TransformFeedbackProgramMismatch` to `DrawError` that is triggered if you use the same session with multiple different programs.
- Add `TransformFeedbackFormatMismatch` to `DrawError` if the vertex buffer doesn't have the right format.

- Add `draw_from_feedback` to the `Surface` trait with takes a `&TransformFeedbackResult` instead of vertices and indices sources.

tomaka:
- Add a method to `Program` to retreive the vertices format.

tomaka:
*Warning*: transform feedback objects aren't shared between contexts.

PR #538, opened 2015-02-20
Implement basic transform feedback

tomaka:
cc #34


tomaka:
I'm getting the following problem: if `DrawParameters` contains a `&mut`, I can't pass it as a `&` or I can't access the content of the `&mut`.

This means that I have to make a breaking change in the `draw` function to pass `DrawParameters` by value.

tomaka:
After some thinking, I'm just going to pass a `&TransformFeedbackSession` instead of a `&mut` one.

PR #539, opened 2015-02-20
External library integration features - expose GlObject and allow in-context execution

ryanstew:
We've been working on exposing the Oculus SDK via FFI to Rust. There's a bit of a handshake with the SDK that isn't possible with the current glium API. Feedback is of course totally welcome on the shape of this; in the end, I'm trying to accomplish two things:

* Expose the OpenGL id of the textures that serve as color attachments for framebuffers. In the case of the Oculus SDK, it's required to do all rendering to a pair of frame-buffers (one per eye), then hand these to the SDK via these ids for further post-processing prior to pushing them out to the display.
* Allow the SDK to run arbitrary code with the glium OpenGL context active. There's quite a bit of OpenGL context manipulation done by the C API itself. Though the SDK does create its own OpenGL context for some of the work it does, there is pre- and postamble work done on your behalf per frame, and init code that also has to use your OpenGL context.

   I know there are already some open issues discussing the longer-term implications of the threading model used by glium currently. This seemed like the most durable approach even if a change is made in the threading/context model.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -224,8 +224,8 @@ mod gl {
     include!(concat!(env!("OUT_DIR"), "/gl_bindings.rs"));
 }
 
-/// Internal trait for objects that are OpenGL objects.
-trait GlObject {
+/// Trait for objects that are OpenGL objects.
+pub trait GlObject {
     type Id;
 
     /// Returns the id of the object.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1113,6 +1113,23 @@ impl Display {
         ops::read_from_default_fb(gl::FRONT_LEFT, self)
     }
 
+    /// Execute an arbitrary closure with the OpenGL context active. Useful if another
+    /// component needs to directly manipulate OpenGL state.
+    ///
+    /// **If action manipulates any OpenGL state, it must be restored before action
+    /// completes.**
+    pub unsafe fn exec_in_context<'a, T, F>(&self, action: F) -> T
+                                            where T: Send + 'static,
+                                            F: FnOnce() -> T + 'a
+    {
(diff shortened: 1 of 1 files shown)


tomaka:
Features are not the right way to do this. Just make the `GlObject` trait always public (and remove `expose_globject`).

src/lib.rs
```diff
@@ -1113,6 +1123,17 @@ impl Display {
         ops::read_from_default_fb(gl::FRONT_LEFT, self)
     }
 
+    /// Execute an arbitrary closure with the OpenGL context active. Useful if another
+    /// component needs to directly manipulate OpenGL state.
+    pub unsafe fn exec_in_context<'a, T: std::marker::Send, F: FnOnce() -> T + 'a>(&self, action: F) -> T {
```
tomaka: While you're at it, this can be just `Send`, no need for the namespace.

src/lib.rs
```diff
@@ -1113,6 +1123,17 @@ impl Display {
         ops::read_from_default_fb(gl::FRONT_LEFT, self)
     }
 
+    /// Execute an arbitrary closure with the OpenGL context active. Useful if another
+    /// component needs to directly manipulate OpenGL state.
```
tomaka: And please add a warning saying that the function must preserve the OpenGL state. Glium will just totally break if this isn't respected.

tomaka:
Could you rebase over master now that the library compiles again? :-/

ryanstew:
Sure thing, been a busy couple days of big breaking changes it seems. I had to make one small update to exec_in_context for yesterday's nightly. Should be building clean now with all your suggestions in place. Let me know if anything else looks off. Thanks!

PR #569, opened 2015-02-23
Fix fbo creation again

tomaka:


src/fbo.rs
```diff
@@ -467,10 +467,27 @@ unsafe fn attach(ctxt: &mut context::CommandContext, slot: gl::types::GLenum,
 
         match attachment {
             Attachment::Texture { id: tex_id, level, layer, .. } => {
-                ctxt.gl.FramebufferTextureLayer(gl::DRAW_FRAMEBUFFER,
-                                                slot, tex_id,
-                                                level as gl::types::GLint,
-                                                layer as gl::types::GLint);
+                match bind_point {
```
palodequeso: looks like this variable doesn't exist in scope

Issue #402, opened 2015-01-29
Disable debug output panics while compiling shaders


tomaka:
Otherwise compilation errors aren't recoverable.

PR #579, opened 2015-02-24
Add circle-ci integration

tomaka:


tomaka:
It looks like it's working, except that #402 makes tests fail.

tomaka:
Note that the results don't show up here because I'm disabling it until tests are passing.

tomaka:
Yay, circle-ci is passing.

Issue #513, opened 2015-02-15
Add get_supported_glsl_version() to Display


tomaka:

tomaka:
Done in #586

PR #586, opened 2015-02-24
Retrieve glsl version from gl version

Futile:
This is meant to implement #513.


palodequeso:
Just curious about which platforms you have tested this on?

Futile:
Not sure what you mean. The implementation only matches on the GlVersion already retrieved by glium, and differentiates between gl and gl es. No open gl calls are performed.

tomaka:
After some thoughts, I'd prefer if you removed `Version::get_glsl_version` and moved its body in `Display::get_supported_glsl_version`.

Since the `Version` struct can both mean an OpenGL version or a GLSL version, it wouldn't make sense to be able to call `get_glsl_version` on a `Version` object that already represents a GLSL version.

Futile:
Yeah, I noticed that aswell. So moving into `Display::get_supported_glsl_version` is preferred over copying `version::Version` and its `PartialOrd` implementation into a new type `GlslVersion`?

Although only of secondary concern, but moving into `Display::get_supported_glsl_version` will also make testing more difficult and less atomic.

tomaka:
You could also write:

```rust
pub struct Version(Api, u8, u8);

impl Version {
    ...
}

// this function is invisible to the outside since the "version" module is private
pub fn gl_version_to_glsl_version(_: &Version) -> Version {
    ...
}

#[cfg(test)]
mod test {
    ...
}
```

tomaka:
You will need to rebase over master so that travis passes.

PR #587, opened 2015-02-25
Improve buffers binding state handling

tomaka:


tomaka:
After re-reading the specs, this change isn't needed.

Issue #332, opened 2015-01-20
Make program::Attribute public and add get_attribute() method to Program


tomaka:
The biggest change is turning the `ty: gl::types::GLenum` field into an `AttributeType`.

tomaka:
Done in #589

PR #589, opened 2015-02-25
Make the "Attribute" struct public

tomaka:
cc #332


tomaka:
The previous implementation was slightly wrong. Hopefully I don't break any code by making this change.

PR #599, opened 2015-02-26
Update for gl_generator

tomaka:
Won't pass before https://github.com/tomaka/glutin/pull/299

PR #600, opened 2015-02-27
Fix build error: Catch up gl_generator API changes

simnalamburt:
It was always `Fallback::All` before, but now you can choose one of `All` and `None`.

Reference: https://github.com/bjz/gl-rs/commit/368255d830ca991b029a3b324650aaff2d2560cf


simnalamburt:
Closed by #599

Issue #553, opened 2015-02-23
Glium without Glutin


mrkishi:
As far as I understand, Glium is currently tied to Glutin. Will we ever be able to use Glium without Glutin?

If so, why does Glium's context wraps, for instance, windows and event polling? How would one integrate different backends to Display/DisplayBuild? Admittedly, I'm no OpenGL expert. Can anyone shed some light on this? Was this design a requirement or will Glium leave those concerns to windowing libraries once Glutin ties are abstracted away?

Thanks!

tomaka:
The reason why glium was tied to glutin is that glium uses a background thread which needed access to the window's functions. Therefore I needed ownership of the window and had to expose functions like `poll_events` or `wait_events`. And to avoid having a generic parameter on the `Display` for the event, I have just decided to use glutin.

I've been doing some changes recently to "un-tie" glutin from the backend. I'm not yet totally sure about what I'm going to do. There are some problems that are thought to solve, for example I need to know when the window's OpenGL context is being rebuilt so that I can clear the framebuffer objects and VAOs.

tomaka:
Here's an idea:

```
#[derive(Clone)]
pub struct Context(...);   // opaque type, contains an Rc

pub trait ToContext {
    fn to_context(&self) -> Context;
}

impl AsContext for Display {
    ...
}
```

All glium functions (like `Texture::new`, `VertexBuffer::new`, etc.) would take a generic `&T where T: AsContext` instead of just a `&Display`. This would maintain backward-compatibility.

It would thus be possible to create multiple Display-like types: `GlutinDisplay`, `SdlDisplay`, etc.

mrkishi:
It seems like this ToX/AsX pattern is becoming very useful in Rust!

A question: Wouldn't the `Display` type still limit the feature-availability of the multiple backends, with the additional work of creating these Display-like types? Wouldn't it make sense to base the future of Glium on `Context` instead, leaving window-related duties to the other libraries?

Take for instance the single-context multi-window use case: isn't that something the current `Display` interface can't represent? Wouldn't one have to deal with the raw `SDL` structures, bypassing `SdlDisplay`, or either re-expose those features through `SdlDisplay`? To me, it feels like Glium would be mixing different responsibilities.

tomaka:
I'm not yet decided on this subject.

I'll leave the choice for now (between taking owernship of the backend or not) by staying retro-compatible, then I'll see what it gives.

There are some though cases, like rebuilding the context, which I'm unsure of.

[...]

PR #610, opened 2015-03-03
[Breaking change] Make everything single-threaded

### LINKED ISSUE #459: Threading model discussion
## How OpenGL works

To do something with OpenGL, there are three steps:

 - Create the context
 - Set the context as being the active context in the current thread
 - Call the `glSomething` functions

Where a function in a regular C library would look like `do_something(const Context* context, int param);`, OpenGL instead looks like `do_something(int param);` and operates on the context that is active in the current thread.

Additional restrictions: you can only have one active context per thread, and the same context can't be active in two different threads at the same time (though this sometimes works depending on the backend).

Also on windows it is **very** slow to change the current context. So we should avoid this too.

## How Glium currently handles this

Glium creates an internal background thread that is inaccessible from outside the library. The context is created by glium and set as current in this background thread. This ensures that the context isn't active anywhere else, and that the current context of the thread is never changed.

Then all function calls add a command to a queue that is executed by this background thread.

This approach has the following drawbacks:

- It is slower than raw OpenGL calls.
- Embedded platforms often can't afford an additional thread.
- Having hidden costs is not what you expect from a system language.

## What's the alternative?

Glium aims to be 100% safe, which is an additional constraint in the design.

There are two problems when you want to remove this background thread:

- We must ensure that we are on the right thread when calling OpenGL functions
- We must ensure that the current context didn't change

The first problem is easily solved by removing `Send` and `Sync` from *all* structs that represent an OpenGL object (the display, textures, buffers, etc.).

This has the disadvantage that glium users must do everything from within the same thread, possibly reimplementing something similar to the commands queue that currently exists in glium.

The second problem is a bit more annoying:

 - The user can call `MakeCurrent` manually, but the function is marked as unsafe so we can ignore this situation.
 - The user can create a second context in the same thread.
 - The user can use another OpenGL wrapper which also exposes a safe interface, and create a context in the same thread.

The only solution that I see to make this safe is calling `*GetCurrentContext()` before each OpenGL function call. I don't know how this affects performances.

## TL;DR

There are two alternatives: 

- Having a background thread (the current one)
- Not implementing Send+Sync and calling `GetCurrentContext()` all the time

tomaka:
Glium types no longer implement `Send` and `Sync`. No other change in the public API.
Should lead to a performances gain and something *closer to the metal*.

Need to add some safety and cleanup the code but I wanted to get some feedback first.

Close #459


tomaka:
cc @palodequeso @TyOverby @XMPPwocky @rsaarelm @andrewrk @nilbold and all people that I know use glium

Is this a problem for you if all objects (`Display`, `Texture2d`, `VertexBuffer`, etc.) no longer implement `Send` and `Sync`?

andrewrk:
I've given up on rust for now and went back to C++ :-(

So I no longer depend on glium.

nilbold:
It shouldn't cause any significant problems with what I'm planning to do.

TyOverby:
Nope, my code is all single-threaded anyway.

dylanede:
So is this leading up to removing the dedicated context thread? I'm personally all for this. The choice of expenses due to multithreading should be left to the user, and they should be left to figure out how to synchronise all other data with the rendering thread. If so required such an abstraction as the one that already exists deep inside glium can be built on top of it instead at a later date.

kmcallister:
Fine by me.

It would be great to have a companion library that provides a safe and high-level way to send textures and other resources between GL contexts.  This is unfortunately a huge, platform-specific pain.

kmcallister:
cc @pcwalton

ryanstew:
Per our discussion on IRC, this sounds like great news to me (particularly with direct exposure of a make_current() API to allow external libs to play inside the OpenGL context).

XMPPwocky:
Wouldn't break anything for me.

tomaka:
Thanks for the feedback!

For safety, there will be a flag in the context:
 - If the flag is set (it is by default), all calls to OpenGL functions will be preceded by `{glX/wgl}GetCurrentContext()` to check if the context is the current one, and `{glX/wgl}MakeCurrent` if necessary.
 - If the flag is clear, this verification won't be done. The only way to set this flag to 0 is through an unsafe API.

tomaka:
I'm going to stop here for now in the PR.
The code is dirty, but implementing this correctly would be a huge amount of rework that would conflict with #553

tomaka:
A new nightly came in-between and broke the build, but the last change is minor and I've successfully tested locally.

I'll merge in a few hours if no-one is opposed to this.

tomaka:
:shipit:

PR #611, opened 2015-03-04
Improve first tutorial to use a loop

tomaka:
It has been reported that *not* using a loop sometimes shows a black screen (probably because draw messages aren't processed).


tomaka:
Build time exceeded in one of the builds.

PR #616, opened 2015-03-05
Cleanup the code after the single threading change

tomaka:
Follow-up of #612 and #610 

Will eventually need another cleaning pass for the style, but that's low priority.


tomaka:
ICE: https://github.com/rust-lang/rust/issues/23068

TyOverby:
I don't think that this PR is causing the ICE.  Even though this PR isn't merged, I'm still ICEing.

PR #622, opened 2015-03-08
[Breaking change] Use dependencies from crates.io

tomaka:
This will allow publishing glium on crates.io.


tomaka:
This is a breaking change, as people who use glium will also need to use the crates.io versions of cgmath, nalgebra and/or image if they want to use these types with glium.

PR #654, opened 2015-03-17
Add Texture::get_format function

### LINKED ISSUE #499: Add `get_format()` for textures
Can be retreived with `glGetTexLevelParameter`

tomaka:
Remarks:

 - Not all formats are implemented, most notably sRGB and compressed formats. This will need to be added in other PRs.
 - OpenGL 1.0 doesn't support `glGetTexLevelParameterfv` ; this may need to be tackled, or maybe not
 - OpenGL ES 2 doesn't support this function either. However GLES2 is not allowed to perform any conversion when you upload data, so the format can be determined like this. This will need be tackled when GLES2 support is added.

Close #499


tomaka:
Too many unanswered questions in the design.

PR #668, opened 2015-03-21
Buffer rework

tomaka:
I'm trying to get things right in `buffer.rs` before making things clean over it.

Internal change only.


tomaka:
Circle-ci fails because of a bug with rustup. Merging.

PR #681, opened 2015-03-26
Rustup

tomaka:


tomaka:
Travis fails because there is no output on stdout for 10mn :(

PR #690, opened 2015-03-28
Calculate correct framebuffer size with hidpi_factor

derkyjadex:
This fixes the incorrect viewport size when running on a retina screen in OS X.

When moving the window between retina/non-retina monitors the viewport is incorrect again, but resizing the window fixes it. I don't know why this is happening because `sync_viewport_scissor` calls `gl.Viewport` with the correct size every time. In any case, this is better than the current situation.


tomaka:
Thanks!

Issue #293, opened 2015-01-16
Cache the values of uniforms


tomaka:

palodequeso:
Do you mean cache them so if you pass the same reference to a uniform struct, it won't call glUniform* again?

tomaka:
Don't call `glUniform` if possible.

palodequeso:
makes sense.

tomaka:
Done in #711

PR #711, opened 2015-04-02
Now caching some uniform values

tomaka:
Pure optimization.

cc #293

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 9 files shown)


tomaka:
Changes are:

 - The state of texture units is now cached in `GLState`.
 - `glBindTexture`/`glBindSampler` are not called if not necessary.
 - The current values of uniforms for scalars, vectors and matrices are now stored in the `Program` struct. `glUniform*` is not called if not necessary.

tomaka:
Gained around 10% performances with this change.

PR #760, opened 2015-04-20
[Small breaking change] Parametrize the type of error in DisplayBuild

tomaka:
- `GliumCreationError` no longer contains a `glutin::CreationError`. Instead it has a template parameter `GliumCreationError<T>` where `T` is the type of the error that can happen when creating the backend.
- The `DisplayBuild` trait has a new associated type `Err` that corresponds to the type of the error.

This PR will make it possible to turn `glutin` into an optional feature.


tomaka:
Do not merge, forgot to change the CHANGELOG.

PR #771, opened 2015-04-23
[Small breaking change] Buffers reform

tomaka:
* Removes the `gl_persistent_mapping` feature.
* Removes `new_persistent` and `new_persistent_if_supported` functions.
* Instead, `VertexBuffer::dynamic` will try to create a persistent buffer, and `VertexBuffer::new` will create an immutable buffer.
* Uniform buffers are always dynamic.
* Index buffers are never dynamic (as there is no write/read/map function available yet).
* The `new_dynamic` function is deprecated in favor of just `dynamic`
* Added `empty_dynamic`
* Moved some tests from "vertex_buffer" to "vertex_source".
* Duplicated all vertex buffer tests for immutable/dynamic.

If you use `VertexBuffer::new` and never modify the buffer, performances will improve. If you use `VertexBuffer::new` and do modify the buffer, performances will be worse, and you should use `VertexBuffer::dynamic` instead. If you use `VertexBuffer::dynamic`, performances will improve.

In any case, the write/read/map operations all still work.


tomaka:
Gained around 3 to 10% performance with immutable buffers after this change.

PR #780, opened 2015-04-26
[Breaking change] Add diagnostic message to assert_no_error.

TyOverby:
This is useful when assert_no_error() fails and you are trying
to find out where in the codebase the error is coming from.

I've been using this change to place assertions in various places
like so:

```rust
display.assert_no_error(Some("error after display creation"));
...
display.assert_no_error(Some("error before draw"));
...

```


tomaka:
Thanks!

PR #783, opened 2015-04-27
Optimize texture units usage

tomaka:
## Current behavior

We start at texture unit 0.
When uniforms are enumerated, each texture is binded to the current texture unit if necessary, and the texture unit is incremented by one.

This means that glium only uses a few texture units and overwrites them very often.

## New behavior

When uniforms are enumerated and a texture is found, glium tries to find a unit where the texture is already binded. If it doesn't find one, it tries to use a new slot (one that has never been used before). If it doesn't find one, it unbinds a random texture.

## Consequence

On code where the same program is used multiple times with different textures, this replaces calls to `glBindTexture` by calls to `glUniform`. On program transitions, this may avoid calls altogether.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -208,6 +208,7 @@ mod image_format;
 mod ops;
 mod sampler_object;
 mod sync;
+mod utils;
 mod version;
 mod vertex_array_object;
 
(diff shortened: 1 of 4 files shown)


tomaka:
This PR also serves as a foundation for caching uniform buffer slots, which will increase performances much more.

PR #791, opened 2015-04-30
examples replace use glium::glutin with extern crate glutin

Hinogary:


tomaka:
That was done on purpose in #785 because I keep getting reports about `build_glium` not implemented on `glutin::WindowBuilder`.

Issue #792, opened 2015-04-30
document the undocumented public items


dcampbell24:
For the sake of less warnings during compilation and documentation itself.

tomaka:
Actually it's on purpose that I didn't remove all warnings for the moment.

The travis-ci test fails if Cargo doesn't write anything on stdout for more than 10 minutes, which is what happens if there is no warning :(
I know you can theoretically fix it with a small utility, but I didn't manage to make it work.

dcampbell24:
Maybe [--progress](https://github.com/rust-lang/rust/issues/24335) will land in the not to distant future and then that can be used instead.

tomaka:
Oh, that would be nice \o/

PR #793, opened 2015-04-30
Eliminate warnings

dcampbell24:
Update glium so that there are no warnings with the latest nightly. Still need to add doc strings #792.

### DIFF
--- a/tests/display.rs
+++ b/tests/display.rs
@@ -350,5 +350,5 @@ fn debug_string() {
     use glium::backend::Facade;
 
     let display = support::build_display();
-    display.get_context().insert_debug_marker("Hello world").unwrap();
+    display.get_context().insert_debug_marker("Hello world").ok();
 }
--- a/tests/display.rs
+++ b/tests/display.rs
@@ -350,5 +350,5 @@ fn debug_string() {
     use glium::backend::Facade;
 
     let display = support::build_display();
-    display.get_context().insert_debug_marker("Hello world");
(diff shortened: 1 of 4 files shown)


dcampbell24:
I added documentation for a few items, but am not very familiar with this library right now, so I think I should defer the rest of the documentation to someone else or a later time.

tests/display.rs
```diff
@@ -350,5 +350,5 @@ fn debug_string() {
     use glium::backend::Facade;
 
     let display = support::build_display();
-    display.get_context().insert_debug_marker("Hello world");
+    display.get_context().insert_debug_marker("Hello world").unwrap();
```
tomaka: The `insert_debug_marker` function isn't guaranteed to work (and it doesn't work on travis/circle-ci).
You should put `display.get_context().insert_debug_marker("Hello world").ok();` instead.

tomaka:
Thanks for the PR!
I'll merge after you fix that problem.

tests/display.rs
dcampbell24: What is the test checking? Should there be another test to see whether `insert_debug_marker` works on the given platform? Maybe we can have some kind of conditional that expects failure on certain platforms?

tomaka: The test is checking that the function exists and doesn't change its API.
You can't predict whether or not it is supposed to succeed, as it depends on the hardware and configuration of the user.

tomaka: More than the API, the test also checks that the function doesn't trigger any OpenGL error.

PR #798, opened 2015-05-02
Added a macro for assert_no_error

TyOverby:
This macro will insert the file and line number into the error message for the `assert_no_error` function call.


tomaka:
I'd prefer `assert_no_gl_error!`, because `assert_no_error!` seems too generic and could conflict with another library or the std.

While you're at it, could you also add a bit of documentation to the macro?

TyOverby:
Renamed to `assert_no_gl_error` and added some basic documentation along with an example that is "ignore"d because I didn't want to have to build a whole display object just to show how to use it.

PR #828, opened 2015-05-09
Add .sampled() methods to textures

tomaka:


tomaka:
I'll also use this PR to fix the compilation on nightly, since it's a minor fix.

Issue #574, opened 2015-02-23
OpenGL ES 2/3 mega-review


tomaka:
The following modules must be reviewed for GLES2 and 3 compatibility:

- [x] buffer (#632)
- [x] context (#629)
- [x] debug (#576)
- [x] draw_parameters (nothing)
- [x] fbo (#650 - more or less hacky, but that's not specific to GLES)
- [x] framebuffer (nothing)
- [ ] image_format
- [x] index (nothing)
- [x] lib
- [ ] ops/blit
- [x] ops/clear (#658)
- [ ] ops/draw
- [x] ops/read
- [x] pixel_buffer (nothing)
- [x] program (partial in #638)
- [x] render_buffer (#584)
- [ ] sampler_object
- [x] sync (#659)
- [x] texture (#773)
- [ ] uniform
- [x] version
- [x] vertex (nothing)
- [ ] vertex_array_object (partial in #628 and #725)

cc #183

tomaka:
I'm going to close this as most things have been reviewed and tests can handle the rest.

PR #858, opened 2015-05-14
[Breaking change] Rework reading from a framebuffer

tomaka:
- Changes `PixelBuffer` to take the type of pixels as template parameter.
- Reworks the content of `ops::read`.
- The `TextureDataSink` traits now take the type of pixel as template parameter.
- The `read` functions now read to a `(u8, u8, u8, u8)`, since this is the only format that is guaranteed to be supported for GLES 2.

cc #574

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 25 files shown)


tomaka:
Will only be merged in a few days/weeks, before 0.5 is released.

PR #864, opened 2015-05-17
remove `use std::default::Default` and `std::default::*`

dcampbell24:
Default is in the prelude so we don't need to import or include the full path.


tomaka:
Thanks!

PR #871, opened 2015-05-19
Implemented Texture2dArray

Esption:
Implements a Vec<RawImage2d> to RawImage3d functions and uses it in the build/textures.rs bit.

This probably isn't the best way, but it works.

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -643,17 +643,11 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
                 ")).unwrap(),   // TODO: panic if dimensions are inconsistent
 
             TextureDimensions::Texture2dArray => (write!(dest, "
-                    let array_size = 0;
-                    let internal_data: Cow<'a, [u8]>  = Cow::Owned(Vec::<u8>::new());
-                    let width = 0;
-                    let height = 0;
-                    let client_format: ClientFormat = unsafe {{ ::std::mem::uninitialized() }};
-                    let mut vec_raw = Vec::new();
-                    for ite in data {{
-                        vec_raw.push(ite.into_raw());
-                    }}
-                    let RawImage3d {{data: internal_data, width, height, depth: array_size, format: client_format }} = RawImage3d::from_vec_raw2d(&vec_raw);
-                    let data = internal_data;
+                    let mut vec_raw = data.into_iter().map(|e| e.into_raw()).collect();
+                    let RawImage3d {{data, width, height, depth: array_size, format: client_format }} = match RawImage3d::from_vec_raw2d(&vec_raw) {{
+                        Ok(var)  => var,
+                        Err(reason) => {{return Err(reason);}}
+                    }};
                 ")).unwrap(),   // TODO: panic if dimensions are inconsistent
 
             _ => unreachable!()
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -643,7 +643,7 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
                 ")).unwrap(),   // TODO: panic if dimensions are inconsistent
 
             TextureDimensions::Texture2dArray => (write!(dest, "
-                    let mut vec_raw = data.into_iter().map(|e| e.into_raw()).collect();
+                    let vec_raw = data.into_iter().map(|e| e.into_raw()).collect();
                     let RawImage3d {{data, width, height, depth: array_size, format: client_format }} = match RawImage3d::from_vec_raw2d(&vec_raw) {{
                         Ok(var)  => var,
                         Err(reason) => {{return Err(reason);}}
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks, that's a good idea in fact!
I can imagine Texture3Ds use this function as well.

build/textures.rs
```diff
@@ -644,11 +644,16 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
 
             TextureDimensions::Texture2dArray => (write!(dest, "
                     let array_size = 0;
```
tomaka: These definitions aren't needed. When you write `let RawImage3d { data, width, etc. } = ...` it creates new variables.

build/textures.rs
```diff
@@ -644,11 +644,16 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
...
-                    let data = Cow::Owned(Vec::<u8>::new());
+                    let internal_data: Cow<'a, [u8]>  = Cow::Owned(Vec::<u8>::new());
                     let width = 0;
                     let height = 0;
-                    let client_format = unsafe {{ ::std::mem::uninitialized() }};
-                    unimplemented!();
+                    let client_format: ClientFormat = unsafe {{ ::std::mem::uninitialized() }};
+                    let mut vec_raw = Vec::new();
```
tomaka: Could you replace this by `data.into_iter().map(|e| e.into_raw())`? It's usually faster because it allocates exactly the right space.

build/textures.rs
```diff
@@ -644,11 +644,16 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
...
+                    for ite in data {{
+                        vec_raw.push(ite.into_raw());
+                    }}
+                    let RawImage3d {{data: internal_data, width, height, depth: array_size, format: client_format }} = RawImage3d::from_vec_raw2d(&vec_raw);
+                    let data = internal_data;
```
tomaka: Just use `RawImage3d { data, width, height, ... }` and remove this line.

src/texture/mod.rs
```diff
@@ -358,6 +358,35 @@ pub struct RawImage3d<'a, T: Clone + 'a> {
...
 }
 
+impl<'a, T: Clone + 'a> RawImage3d<'a, T> {
+    ///Transforms a Vec<RawImage2d> into a RawImage3d
+    pub fn from_vec_raw2d(arr: &Vec<RawImage2d<'a, T>>) -> RawImage3d<'a, T> {
+        let depth   = arr.len();
+        let width   = arr[0].width;
+        let height  = arr[0].height;
+        let format  = arr[0].format;
+        let clo = |v, w, h, d, f| {
```
tomaka: Why not just `let data = { ... };` instead of using a closure?

Esption:
On thinking about it a tad more, it should probably return a ```Result<RawImage3d>``` and then return ```Err``` instead of aborting.  Will do that quickly.

src/texture/mod.rs
Esption: I think that has to do with me spending too much time on this, and not noticing that after the fact.

build/textures.rs
Esption: I didn't know, the high level stuff for vec isn't something I know how to use as of now.

build/textures.rs
Esption: Those were there before hand, and I've never seen the ``let {data, width, ...} = RawImage3d;``` syntax get used before, so thought it was needed.  Learned something new today at least :)

src/texture/mod.rs
```diff
@@ -358,6 +358,38 @@ pub struct RawImage3d<'a, T: Clone + 'a> {
...
+    pub fn from_vec_raw2d(arr: &Vec<RawImage2d<'a, T>>) -> Result<RawImage3d<'a, T>, TextureMaybeSupportedCreationError> {
+        let depth   = arr.len();
+        let width   = arr[0].width;
+        let height  = arr[0].height;
+        let format  = arr[0].format;
+        let raw_data = {
+            let mut vec = Vec::<T>::with_capacity((width * height * depth as u32) as usize);
+            for i in arr {
+                if width != i.width || height != i.height {
+                    return Err(TextureMaybeSupportedCreationError::CreationError(TextureCreationError::DimensionsNotSupported));
```
tomaka: I preferred the panic! Same below.

Esption: You sure?  It seems a bit nicer to have an error than a panic to me.  Care to explain why?

tomaka: The error here describes dimensions and formats that are not supported by the hardware, so that's irrevelant here.

The user is supposed to pass images that have identical dimensions. If it's not the case, it's clearly the user's fault.

Esption: Ah, okay.  The exact errors aren't something I knew the reasons of (only wrote this because I was converting raw opengl calls to glium, and noticed this wasn't supported)

I hadn't realised you wanted the user to be the one to enforce consistent dimensions, not glium.  Makes sense though.

tomaka:
Looks good now, I'll merge after tests pass :)

PR #876, opened 2015-05-20
Doc comments for feature'd functions in build/textures.rs

Esption:
As brought up in https://github.com/tomaka/glium/issues/874


tomaka:
Thanks

PR #880, opened 2015-05-20
Implemented Texture1dArray

Esption:
Basically just a slight edit from the code for the Texture2dArray.

This should also close https://github.com/tomaka/glium/issues/138


tomaka:
Thanks

Issue #675, opened 2015-03-24
Make glutin an optional dependency


tomaka:
Still active by default.

tomaka:
`GliumCreationError` still depends on glutin. Needs further changes before doing that.

PR #898, opened 2015-05-22
Glutin as an optional dependency.

Esption:
For #675

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
     - GLIUM_HEADLESS_TESTS=1
   matrix:
     - FEATURES='' TEST=0
-    - FEATURES='headless' TEST=1
+    - FEATURES='headless glutin' TEST=1
     - FEATURES='headless glutin image cgmath nalgebra gl_read_buffer gl_depth_textures' TEST=1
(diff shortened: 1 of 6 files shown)


tomaka:
This requires changing the `.travis.yml` to add `glutin` in `FEATURES`.
Please also add a line with `FEATURES='' TEST=0`

Esption:
Hope I did that correctly.  Never seen a Travis file before.

.travis.yml
```diff
@@ -14,9 +14,10 @@ env:
     - secure: "AfE+PLANOT0mgUKpgGUu4yIrSNKxoNST5zFBeajONEEE0wl/bZVE6iGqucNPR0EaIgMfk+NGgCLLM7VFfy+HF5alBiqUUOapef+fc91iHVzoBgtlLXOO7E1UntrieBQCodURTaBtotCMe7jtlkNajkvUD6atKTYh1AYJ3JY2bHA="
     - GLIUM_HEADLESS_TESTS=1
   matrix:
+    - FEATURES='' TEST=0
     - FEATURES='headless' TEST=1
```
tomaka: Here as well, or the tests won't work.

Esption: Wouldn't it be better to make the tests that rely on glutin only work if glutin is enabled? I had forgotten that would be needed, but that shouldn't be too hard.

tomaka: Well, yeah. 99% of the tests create a headless context and run OpenGL commands. If you remove glutin, there's nothing left to create the context.

tomaka:
Ah, circle.yml must also be modified to pass `--features "headless glutin"` instead of just `--features "headless"`.

tomaka:
Thanks

PR #906, opened 2015-05-25
Add a fallback if the user forgets a buffer Mapping

tomaka:


tomaka:
Hum. Index buffers and pixel buffers are problematic because they haven't been switched to the "new" buffer design yet.

tomaka:
Will merge after 0.5

PR #914, opened 2015-05-26
Bump versions of dependencies

tomaka:


tomaka:
Apparently there's a stack overflow with Rust 1.0 that doesn't happen with the nightlies.

tomaka:
No longer relevant.

PR #921, opened 2015-05-30
Feature flag doc comments

Esption:
* Standardized to just `# Features` -- A couple weren't, but the majority were.  There doesn't appear to be any differences on the website that I could tell, but it seems silly for it to be different anyway.
* Fixed the double quotes from another commit I did in `build/textures.rs` to be single quotes.
* Added some missing comments about feature flags being necessary.


tomaka:
Thanks!

PR #943, opened 2015-06-05
Add event processing to tutorial loops

TimNN:
Add the line `for _ in display.poll_events() {}` in each main loop of the tutorial code. Without this line the window is unresponsive (on OS X).


tomaka:
Thanks!
However this needs some more changes. I think the code should just call `poll_events()` and process the `Closed` event instead of calling `is_closed()`. And the explanation text in the tutorial should thus be changed as well to explain that.

PR #944, opened 2015-06-07
[fix] framebuffer dimensions for non int scales

rozaliev:
It was broken for scales like ~2.608 (iphone 6 plus) producing x2 size.


tomaka:
Thanks!

Issue #197, opened 2015-01-02
Implement the Attribute trait for cgmath and nalgebra types


tomaka:
Tests should be added to guarantee the memory layout.

thelink2012:
@tomaka Though it's still missing the Attribute trait for nalgebra, I don't think this should be closed?

PR #966, opened 2015-06-10
Attribute trait for cgmath types

thelink2012:
cc #197

**Note:** The layout tests mayn't be that necessary (for cgmath nor for nalgebra) given the fact they have methods that transmute between arrays and their types.


tomaka:
Thanks!

Issue #955, opened 2015-06-09
Implement and/or test arrays and structs in uniform buffers and SSBOs


tomaka:

tomaka:
Fixed by #985

PR #985, opened 2015-06-13
Correct uniform block and shader storage block layouts

tomaka:


tomaka:
I'm going to delay this change because I'm not sure of the actual layout thingy.

tomaka:
Should be good now.

tomaka:
So what this does is that it replaces the old clumbsy `Vec<UniformBlockMember>` with a `BlockLayout` struct. This struct can represent any possible layout in an SSBO or a uniform block.

The next step is to add a `Layout` trait or something similar that can be implemented on structs just like `Vertex` and that can match a `BlockLayout`.

tomaka:
I think I'm going to merge this now without bumping to `0.7`.
This change only affects people who were explicitely referring to the `UniformBlockMember` struct, which in my estimates is approximately nobody.

tomaka:
Fix #955

PR #986, opened 2015-06-14
Uploading and downloading compressed texture data

### LINKED ISSUE #326: Allow reading and re-uploading compressed texture data
It is possible with OpenGL to upload textures that are already compressed and to read compressed texture without decompressing them.

thelink2012:
Hopefully closes #326
**Don't merge** yet, things still left to do, I'll squash my commits later if necessary.

Please check if the approach I'm taking on this is the correct way to go!?!

Also a few questions I have came up with on the process:
 + At _std/ops/read.rs_ there is a function (`client_format_to_gl_enum`) that kinda of looks like a elder duplicate of one in _image_format.rs_?
 + Why does not `new_texture` seems to allow the generation of mipmaps for compressed textures? [This line](https://github.com/tomaka/glium/blob/db6fc9fce201999dc5996575a0a4f38a4ab0b349/src/texture/any.rs#L127).
 + The auto-generated method `with_format` is catching `NotSupported` as unreachable instead of `CreationError`, I suppose this is not correct?

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -168,6 +168,19 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
         TextureType::DepthStencil => "TextureFormatRequest::AnyDepthStencil",
     };
 
+    // whether this is a internally compressed texture object
+    let is_compressed = match ty {
+        TextureType::Compressed |
+        TextureType::CompressedSrgb => true,
+        _ => false,
+    };
+
+    let client_format_any_ty = match ty {
+        TextureType::Compressed => "ClientFormatAny::CompressedFormat",
+        TextureType::CompressedSrgb => "ClientFormatAny::CompressedSrgbFormat",
+        _ => "ClientFormatAny::ClientFormat",
+    };
+
     // the `#[cfg]` attribute for the related cargo feature
     let cfg_attribute = {
         let format = match ty {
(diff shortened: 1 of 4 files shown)


tomaka:
Wow, thanks for tackling this!

I'm afraid that your approach would lead to too many corner cases:
 - Reading in a compressed client format a texture with a different format would be forbidden (for the moment it's possible to read any texture with any format).
 - Creating a non-compressed texture with a compressed client format.
 - Creating a compressed texture of a specific compression with the wrong compressed client format.
 - The `PixelValue` trait (which represents a single pixel) shouldn't be able to return a compressed data type.
 - etc.

There's a big point that differenciate non-compressed data from compressed data: you can convert from/to any non-compressed format to any other format (when reading/writing to a texture) but you can't do the same with compressed data.

In my opinion it would be better to differenciate functions that operate with compressed data and functions that operate with non-compressed data :
 - Add the `Texture*dCompressedDataSource` traits with a method that returns a `CompressedFormat`, a mipmaps count and some data (this `CompressedFormat` can be the same as the one in `image_format`).
 - Add a constructor function `with_compressed_data` to compressed textures.
 - Add `read_compressed_data` to compressed textures.
 - The `new_texture` function can do the checks at runtime and panic if things are incorrect. Since the function is private, things being incorrect means that there's a bug in glium.

> At std/ops/read.rs there is a function (client_format_to_gl_enum) that kinda of looks like a elder duplicate of one in image_format.rs?

That's correct. I have changed this at some point and forgot to fix this.
They do slightly different things, but it could be possible to make the one in image_format.rs use the one in read.rs (and move the one in read.rs in image_format.rs for a cleaner organization).

> The auto-generated method with_format is catching NotSupported as unreachable instead of CreationError, I suppose this is not correct?

That's in fact correct. `NotSupported` means that the texture type (not the format) is not supported. For example if you try to create a 3D texture with OpenGL ES 2, you will get this.
When a format is not supported, it returns `CreationError::UnsupportedFormat`.

> Why does not new_texture seems to allow the generation of mipmaps for compressed textures? This line.

For the moment mipmaps can only be auto-generated by glium (I know this is a problem).
Calling `glGenerateMipmaps` is forbidden with compressed textures, so if I didn't have this line it would end up generating an OpenGL error.

In my opinion this should eventually be changed to give the user four possibilities:
 - Give data for mipmaps in the `Texture*dDataSource` traits.
 - Auto-generate mipmaps.
 - Only upload the main texture and create empty mipmaps.
 - No mipmaps.

However the API would need some thinking.

But you don't need to do this in your PR. It can be changed separately (just add an `unimplemented!()` if you find a corner case).

thelink2012:
Ready to merge when ci gives green light :+1: 

I've tested only DXT1C since that's the resource I could easily get, I'll be using DXT3/DXT5 very soon so if anything is broken I'll find out (though I don't see why it'd be broken, but you never know). I have not tested RGTC and BPTC in any form since I'm not quite familiar with them (and my integrated card mostly like don't like them).

tomaka:
Nice work!

PR #1001, opened 2015-06-20
[0.6] Allow pre-calculated mimaps

### LINKED ISSUE #997: Allow passing pre-calculated mipmaps
Quoting @tomaka in a random PR:
> For the moment mipmaps can only be auto-generated by glium (I know this is a problem).
Calling glGenerateMipmaps is forbidden with compressed textures, so if I didn't have this line it would end up generating an OpenGL error.
>
>In my opinion this should eventually be changed to give the user four possibilities:
>
>   * Give data for mipmaps in the Texture*dDataSource traits.
>   * Auto-generate mipmaps.
>   * Only upload the main texture and create empty mipmaps.
>   * No mipmaps.
>
> However the API would need some thinking.

thelink2012:
Closes #997

**Breaking changes:**
 + `with_compressed_data` and `with_compressed_data_if_supported` has an additional parameter for mipmaps.
 +  All the texture constructors that used to take a boolean as parameter for mipmaps now takes an enum.

**Other changes:**
+ `empty_with_format` and `empty_with_format_if_supported` are now allowed for compressed textures.
+ `write`, `write_compressed_data`, `write_compressed_data_if_supported`, `read_compressed_data` are now available for mipmap objects. _(No ideia how to implement read)_


tomaka:
Nice, thanks.

PR #1011, opened 2015-06-21
Update to gl_generator 0.0.27

tomaka:
gl_generator 0.0.27 can call glGetError() after each OpenGL error


tomaka:
> thread 'rustc' has overflowed its stack

:(

tomaka:
https://github.com/rust-lang/rust/issues/26467

Issue #998, opened 2015-06-18
panicked during initialization


ConnyOnny:
Hi, I used the exact example code to show a blank window from here: http://tomaka.github.io/glium/glium/index.html
This is what I get:
`thread '<main>' panicked at 'glium has triggered an OpenGL error during initialization. Please report this error: https://github.com/tomaka/glium/issues', /home/c/.cargo/registry/src/github.com-1ecc6299db9ec823/glium-0.5.5/src/context/mod.rs:188`
I'm on an Intel IGPU running Debian Testing with a compositing window manager under Xorg and other OpenGL applications (even games) work fine.

tomaka:
Hi. What does `glxinfo` say?

ConnyOnny:
Works now. Thank you for the fix.

PR #1020, opened 2015-06-22
Fix several OpenGL errors

kornholi:
Fixed several potential OpenGL errors, including a panic during initialization on Intel integrated GPUs (#998).

### DIFF
--- a/src/context/capabilities.rs
+++ b/src/context/capabilities.rs
@@ -277,11 +277,11 @@ pub unsafe fn get_capabilities(gl: &gl::Gl, version: &Version, extensions: &Exte
         },
 
         max_indexed_transform_feedback_buffer: {
-            if version >= &Version(Api::Gl, 3, 0) || extensions.gl_arb_transform_feedback3 {      // TODO: make sure that GL 3.0 supports it   // TODO: GLES
+            if version >= &Version(Api::Gl, 4, 0) || extensions.gl_arb_transform_feedback3 {      // TODO: GLES
                 let mut val = mem::uninitialized();
                 gl.GetIntegerv(gl::MAX_TRANSFORM_FEEDBACK_BUFFERS, &mut val);
                 val
-            } else if extensions.gl_ext_transform_feedback {
+            } else if version >= &Version(Api::Gl, 3, 0) || extensions.gl_ext_transform_feedback {
                 let mut val = mem::uninitialized();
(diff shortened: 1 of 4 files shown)


tomaka:
Oh nice! Thanks!

src/context/capabilities.rs
```diff
@@ -266,16 +266,18 @@ pub unsafe fn get_capabilities(gl: &gl::Gl, version: &Version, extensions: &Exte
...
+                gl.GetIntegerv(gl::MAX_SHADER_STORAGE_BUFFER_BINDINGS, &mut val);
+                val
+            } else {
+                0
+            }
         },
 
         max_indexed_transform_feedback_buffer: {
-            if version >= &Version(Api::Gl, 3, 0) || extensions.gl_arb_transform_feedback3 {      // TODO: make sure that GL 3.0 supports it   // TODO: GLES
+            if version >= &Version(Api::Gl, 4, 0) || extensions.gl_arb_transform_feedback3 {      // TODO: GLES
```
tomaka: Could you replace the `} else if extensions.gl_ext_transform_feedback {` right below with `} else if  version >= &Version(Api::Gl, 3, 3) || extensions.gl_ext_transform_feedback {` ?

tomaka: And I guess we should add `} else if version >= &Version(Api::Gl, 3, 0) { 1 }` as well. Transform feedback exists since OpenGL 3.0 but there's apparently no way to query the number of possible buffers.

tomaka: Ah, nevermind. Forget the previous comment. The `MAX_TRANSFORM_FEEDBACK_SEPARATE_ATTRIBS` in fact appeared in OpenGL 3.0.
So could you replace `} else if extensions.gl_ext_transform_feedback {` with `} else if version >= &Version(Api::Gl, 3, 0) || extensions.gl_ext_transform_feedback {` ?

kornholi: Are you sure it appears in 3.0? I couldn't find it anywhere in https://www.opengl.org/registry/doc/glspec30.20080923.pdf, but it is in the 3.1 specification.

kornholi: Nevermind. For some reason the OS X reader couldn't find it in that file.

tomaka: Page 66.

> The total  number  of  variables  that  may  be  captured  in  separate  mode  is  given  by MAX_TRANSFORM_FEEDBACK_SEPARATE_ATTRIBS.

kornholi: Done!

PR #1025, opened 2015-06-23
Fix error an initialization

tomaka:
cc @meh


meh:
Confirm the error message is no more.

Issue #1026, opened 2015-06-24
Handle per-program point size


tomaka:
- Add a ~~parameter~~ member variable to `Program` stating whether or not it sets `gl_PointSize`.
 - Call `glEnable(GL_PROGRAM_POINT_SIZE)` when using a program where this is the case.

PR #1029, opened 2015-06-24
Add support for GL_PROGRAM_POINT_SIZE

meh:
For #1026.

### DIFF
--- a/src/program/program.rs
+++ b/src/program/program.rs
@@ -29,6 +29,7 @@ use vertex::VertexFormat;
 /// A combination of shaders linked together.
 pub struct Program {
     raw: RawProgram,
+    uses_point_size: bool,
 }
 
 impl Program {
@@ -38,10 +39,11 @@ impl Program {
     {
         let input = input.into();
 
-        let raw = match input {
+        let (raw, uses_point_size) = match input {
             ProgramCreationInput::SourceCode { vertex_shader, tessellation_control_shader,
                                                tessellation_evaluation_shader, geometry_shader,
-                                               fragment_shader, transform_feedback_varyings } =>
+                                               fragment_shader, transform_feedback_varyings,
+                                               uses_point_size } =>
             {
                 let mut has_geometry_shader = false;
                 let mut has_tessellation_shaders = false;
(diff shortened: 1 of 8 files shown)


meh:
Need feedback on the presence of the getter/setter methods for `point_size`, and on the naming of the fields and methods.

src/program/program.rs
```diff
@@ -84,16 +86,18 @@ impl Program {
                     shaders_store
                 };
 
-                try!(RawProgram::from_shaders(facade, &shaders_store, has_geometry_shader,
-                                              has_tessellation_shaders, transform_feedback_varyings))
+                (try!(RawProgram::from_shaders(facade, &shaders_store, has_geometry_shader,
+                                               has_tessellation_shaders, transform_feedback_varyings)),
+                uses_point_size)
```
tomaka: Ugly alignment here

src/program/program.rs
```diff
@@ -262,6 +267,14 @@ impl Program {
     pub fn get_shader_storage_blocks(&self) -> &HashMap<String, UniformBlock> {
         self.raw.get_shader_storage_blocks()
     }
+
+    pub fn uses_point_size(&self) -> bool {
+      self.uses_point_size
+    }
+
+    pub fn set_point_size_usage(&mut self, value: bool) {
```
tomaka: I don't think it should have a setter. That's something that never changes.

tomaka:
Could you:
 1 - Remove the setter.
 2 - Add a comment at `ProgramExt::use_program` that says that it enables/disables GL_POINT_SIZE.
 3 - Add a check in `Program::new` to check that this option is supported by the backend, and a new variant in `ProgramCreationError`.

tomaka:
4 - Add a line in the CHANGELOG

meh:
1 - I thought about the setter for the `Binary` variant, or would there be another way to set it for it? Otherwise should I remove the field from that variant and just default to false?

3 - what should the check be? I keep failing at finding minimum version for stuff, teach me your ways :panda_face:

src/program/program.rs
meh: How should I align it? Just add the missing space to indent beyond the `(`?

tomaka:
To find the minimum version I just go on docs.gl and open each version one by one.
When I'm not sure (there are some errors on this site) I take a look at the official specs.

Here it's minimum OpenGL 3.0, and OpenGL ES doesn't support it.

meh:
Updated.

tomaka:
Thanks!

PR #1034, opened 2015-06-24
Handle GL_EXT_buffer_storage

tomaka:


tomaka:
> type `&gl::Gl` does not implement any method in scope named `BufferStorageEXT`

This is weird. Need to investigate why.

tomaka:
Ah it's because `khronos_api` is too old.

PR #1048, opened 2015-06-30
WIP - Support for matrix attributes

camjackson:
This isn't actually working, I'm not sure why. Just wanted to see if I'm on the right track, and if there are any obvious errors :)

### DIFF
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -456,10 +456,10 @@ unsafe fn bind_attribute(ctxt: &mut CommandContext, program: &Program,
             None => continue
         };
 
-        let (attribute_ty, _) = vertex_binding_type_to_gl(attribute.ty);
-
         if attribute.location != -1 {
+            let (attribute_ty, _) = vertex_binding_type_to_gl(attribute.ty);
             match attribute_ty {
+                //The first three cases also account for Vecs of ints/doubles/floats
                 gl::BYTE | gl::UNSIGNED_BYTE | gl::SHORT | gl::UNSIGNED_SHORT |
                 gl::INT | gl::UNSIGNED_INT =>
                     ctxt.gl.VertexAttribIPointer(attribute.location as u32,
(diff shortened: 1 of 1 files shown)


src/vertex_array_object.rs
```diff
@@ -467,15 +467,36 @@ unsafe fn bind_attribute(ctxt: &mut CommandContext, program: &Program,
...
-                gl::DOUBLE_MAT2 | gl::DOUBLE_MAT3 | gl::DOUBLE_MAT4 |
-                gl::DOUBLE_MAT2x3 | gl::DOUBLE_MAT2x4 | gl::DOUBLE_MAT3x2 |
-                gl::DOUBLE_MAT3x4 | gl::DOUBLE_MAT4x2 | gl::DOUBLE_MAT4x3 =>
+                gl::DOUBLE =>
                     ctxt.gl.VertexAttribLPointer(attribute.location as u32,
                                                  elements_count as gl::types::GLint, data_type,
                                                  stride as i32,
                                                  (buffer_offset + offset) as *const libc::c_void),
 
+                gl::FLOAT => {
```
camjackson: Execution gets here for `vec3`s, and rendering works.

tomaka:
You need to call `glEnableVertexAttribArray` for each value of `attribute.location` as well.

src/vertex_array_object.rs
```diff
@@ -467,15 +467,36 @@ unsafe fn bind_attribute(ctxt: &mut CommandContext, program: &Program,
...
+                    ctxt.gl.VertexAttribPointer(attribute.location as u32,
+                                                elements_count as gl::types::GLint, data_type, 0,
+                                                stride as i32,
+                                                (buffer_offset + offset) as *const libc::c_void)
+                },
+
+                gl::FLOAT_MAT4 => {
```
camjackson: Execution gets here for a `mat4`, and I tried to match up the arguments to `VertexAttribPointer` so that they look like the ones for `gl::FLOAT`. But it still doesn't render.

camjackson:
It works! Sort of.

I added `ctxt.gl.EnableVertexAttribArray(i);`, and it still doesn't render properly, but it does show something now.

tomaka:
Well if you're doing `EnableVertexAttribArray(i)`, you need to do `EnableVertexAttribArray(attribute.location + i)` instead. Not sure if that was a typo or a bug in your code.

camjackson:
Ah, I was just passing `i`. I fixed that, but it's still rendering wrong. Just kinda weird lines all over the place.

It seems like the final `vec3` (which has the translation in it) is the only one that doesn't work.

tomaka:
And are you calling `glVertexAttribDivisor` correctly as well? This one is required for instancing.

camjackson:
It works!

If it wasn't already obvious, I have no idea what I'm doing when it comes to OpenGL haha. But I'm learning :)

I'll clean up the code that I've got and push it hopefully in the next couple of days.

camjackson:
The diff has gotten a little messy, but I think this works now, and the same code handles both vectors and matrices. There's still some duplication in arranging the arguments to the three different `VertexAttrib` functions, I think I might be able to clean those up too.

Do you think the change is OK in principle?

tomaka:
Yeah, it looks good.
Could you add a line in the CHANGELOG as well?

camjackson:
See #1058

Issue #878, opened 2015-05-20
Approaching zero driver overhead


tomaka:
- [x] Multiple "buffers per buffer" (#875)
- [x] Persistent mapping
- [ ] Multidraw indirect (#244)
- [x] Uniform buffer
- [ ] SSBO (#362)
- [ ] Texture arrays (implemented but untested)
- [ ] Bindless textures (#243)
- [ ] Sparse textures
- [ ] Drawing with a base instance

Esption:
Is there something you want done to prove the Texure2dArray is working? The 1dArray isn't implemented at all yet, though.

I quickly hacked up the image example to show it's working. https://github.com/Esption/glium/tree/array_example

I'm not sure you what to add as a second image to show the array part off, so maybe that's bad.  But, it is working.

tomaka:
Let's check texture arrays then :o)

PR #1057, opened 2015-07-02
Bindless textures

### LINKED ISSUE #243: Use bindless textures if possible
With GL_ARB_bindless_texture

tomaka:
Summary of the changes:

 - The buffer types no longer require their content to `Send` or `'static`, except for buffers types which don't have type information.
 - Add the `TextureHandle<'a>` type that can be put into uniform buffers/SSBOs. Vertex buffers and regular uniforms are not yet supported.
 - A texture can be turned into a `ResidentTexture` by calling `resident` or `resident_if_supported`.
 - A `TextureHandle<'a>` can be built from a `&'a ResidentTexture` and can't outlive the resident texture.

The implementation should be entirely safe.

Future work:
 - Add bindless textures support for vertex buffers and regular uniforms.
 - Support using a sampler.
 - Image load/store.
 - Consider strong typing, with `ResidentTexture2dArray` and `Texture2dArrayHandle` for example.

Close #243 
cc #878

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
     - FEATURES='' TEST=0
     - FEATURES='headless glutin' TEST=1
     - FEATURES='headless glutin image cgmath nalgebra gl_read_buffer gl_depth_textures' TEST=1
-    - FEATURES='headless glutin image cgmath nalgebra gl_read_buffer gl_depth_textures gl_read_buffer gl_uniform_blocks gl_sync gl_program_binary gl_tessellation gl_instancing gl_integral_textures gl_depth_textures gl_stencil_textures' TEST=0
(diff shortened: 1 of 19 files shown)


tomaka:
Someone made a good remark on IRC. If `ResidentTexture` takes ownership of the texture, that means we can't use the same texture with multiple different samplers. This is bad.

tomaka:
There are two solutions:
 - Either make `ResidentTexture` take a reference to a texture instead of taking ownership of it.
 - Or make `ResidentTexture` hold multiple handles, one per sampler. This would require changing `TextureHandle::set(&ResidentTexture)` to `TextureHandle::set(&ResidentTexture, &SamplerBehavior)`.

I think that the second option is much easier to use for the user.

PR #1058, opened 2015-07-02
Support for matrix attributes

### LINKED ISSUE #1046: Can't pass a per-instance matrix to a shader
I'm not sure whether this is a bug in Glium, or just me getting something wrong - probably the latter.

I want my shader to take a view matrix (and eventually a projection matrix) as a uniform, and a model matrix that's constant per instance. At the moment I can get it to work if I specify the model translation and scaling as separate `vec3`s, and construct the model transform matrix in GLSL, but I can't seem to get it to work when I construct the model transform matrix in Rust and pass it in as a single `mat4`.

My shader right now:

```glsl
#version 140

uniform mat4 view_transform; //works

in vec3 vertex_position; //works

in vec3 model_position; //works
in vec3 model_scale; //works
in mat4 model_transform; //does not work :(

in vec3 vertex_color;
out vec3 vColor;

void main() {
    //building the model transform from its 2 parts - this works
    mat4 m_t = mat4(
            vec4(model_scale.x, 0.0, 0.0, 0.0),
            vec4(0.0, model_scale.y, 0.0, 0.0),
            vec4(0.0, 0.0, model_scale.z, 0.0),
            vec4(model_position, 1.0)
    );
    gl_Position = view_transform * m_t * vec4(vertex_position, 1.0); //this renders correctly

    //If I comment out the line above and use this one, it renders nothing at all
    //gl_Position = view_transform * model_transform * vec4(vertex_position, 1.0);

    vColor = vertex_color;
}
```
The code that's generating the instances looks like this:

```rust
fn instances(&self, transforms: &Vec<Transform>) -> VertexBufferAny {
    #[derive(Copy, Clone)]
    struct ModelTransform {
        model_transform: [[f32; 4]; 4], //Is this the right type for a mat4? That's what the uniforms use
        model_position: [f32; 3],
        model_scale: [f32; 3],
    }

    implement_vertex!(ModelTransform, model_transform, model_position, model_scale);

    let mut data = Vec::new();
    for t in transforms.iter() {
        if t.visible {
            data.push(ModelTransform {
                model_transform: [     //I want to construct the matrix here, not in GLSL, but it doesn't work
                    [t.scale_x, 0., 0., 0.],
                    [0., t.scale_y, 0., 0.],
                    [0., 0., t.scale_z, 0.],
                    [t.x, t.y, t.z, 1.],
                ],
                //I want to delete the next 2 lines, but at the moment I need them
                model_position: [t.x, t.y, t.z],
                model_scale: [t.scale_x, t.scale_y, t.scale_z],
            })
        }
    }
    VertexBuffer::new(&self.display, data).into_vertex_buffer_any()
}
```

The draw call is something like:

```rust
frame.draw((&vertices, instances.per_instance_if_supported().unwrap()),
    &indices, &program, &uniforms, &std::default::Default::default()
).unwrap();
```

I searched through all the Glium examples, and it turns out there's no example of a non-uniform matrix being passed to a shader. I wouldn't have thought that passing a matrix would be different depending on whether it's a uniform or not, but I can't figure out what I'm doing wrong?

camjackson:
Fixes #1046

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
  - Buffers with persistent mapping now synchronize individual segments of the buffer instead of the buffer as a whole.
  - Dynamic sized arrays are now properly handled by the retrospection system.
+ - Added support for matrix shader attributes
 
 ## Version 0.6.5 (2015-06-29)
 
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -350,58 +350,58 @@ impl GlObject for VertexArrayObject {
     }
 }
 
-fn vertex_binding_type_to_gl(ty: AttributeType) -> (gl::types::GLenum, gl::types::GLint) {
+fn vertex_binding_type_to_gl(ty: AttributeType) -> (gl::types::GLenum, gl::types::GLint, gl::types::GLint) {
     match ty {
-        AttributeType::I8 => (gl::BYTE, 1),
-        AttributeType::I8I8 => (gl::BYTE, 2),
-        AttributeType::I8I8I8 => (gl::BYTE, 3),
-        AttributeType::I8I8I8I8 => (gl::BYTE, 4),
-        AttributeType::U8 => (gl::UNSIGNED_BYTE, 1),
-        AttributeType::U8U8 => (gl::UNSIGNED_BYTE, 2),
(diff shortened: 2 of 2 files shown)


tomaka:
Thanks!

Issue #1056, opened 2015-07-02
Add support for U64 attributes


tomaka:

camjackson:
How about I64?

tomaka:
Sure, I64 as well.

The problem with U64/I64 is that they are supported only by modern cards. For the moment we don't check whether the backend supports vertex attributes. This hasn't been a problem so far, but it will become one if we add U64/I64.

tomaka:
Closing in favor of #1077

PR #1061, opened 2015-07-03
WIP - support for U64 and I64 attributes

camjackson:
I've basically just done the copy-pasta bits. I'm not sure if I've missed anywhere that they need to be listed, and I have no idea how the checking for GPU support would be done. cc: #1056


src/vertex_array_object.rs
```diff
@@ -376,6 +376,14 @@ fn vertex_binding_type_to_gl(ty: AttributeType) -> (gl::types::GLenum, gl::types
...
         AttributeType::U32U32U32 => (gl::UNSIGNED_INT, 3, 1),
         AttributeType::U32U32U32U32 => (gl::UNSIGNED_INT, 4, 1),
+        AttributeType::I64 => (gl::INT, 1, 1),
+        AttributeType::I64I64 => (gl::INT, 2, 1),
+        AttributeType::I64I64I64 => (gl::INT, 3, 1),
+        AttributeType::I64I64I64I64 => (gl::INT, 4, 1),
+        AttributeType::U64 => (gl::UNSIGNED_INT, 1, 1),
+        AttributeType::U64U64 => (gl::UNSIGNED_INT, 2, 1),
+        AttributeType::U64U64U64 => (gl::UNSIGNED_INT, 3, 1),
+        AttributeType::U64U64U64U64 => (gl::UNSIGNED_INT, 4, 1),
```
camjackson: I'm not sure what `gl::` type should be used here, but I'm pretty sure it's not these ones. I tried `gl::LONG` and `gl::UNSIGNED_LONG` but they weren't right. I search through the gl_bindings.rs file in the target directory, and couldn't figure it out.

tomaka:
This is done with https://www.opengl.org/registry/specs/NV/vertex_attrib_integer_64bit.txt (for I64 and U64) and/or https://www.opengl.org/registry/specs/ARB/bindless_texture.txt (U64 only)

It also looks more complicated than regular attributes.

tomaka:
Closing in favor of #1096

PR #1062, opened 2015-07-03
[0.7] Reworking uniforms

tomaka:


tomaka:
I'm still struggling to find a good design, so this PR will likely not be merged for some time.
The "correct ways" are all impossible because of limitations in Rust.

tomaka:
Now rebased over #1070

PR #1066, opened 2015-07-05
[acceptable breaking change] Hold the dimensions in the TextureType

tomaka:
This change cleans things up, and is mostly needed to support cubemaps.
(handling cubemaps without this PR would probably result in a giant spaghetti of code impossible to understand)

Technically a breaking change, but it's highly unlikely that people are using `TextureType` in their code. This PR can be published without bumping to `0.7`.


tomaka:
Actually I'm going to rename `TextureType` to `Dimensions`.

PR #1091, opened 2015-07-14
[Breaking change] Remove the compatibility-related Cargo features

### LINKED ISSUE #1073: Remove the compatibility system with Cargo features
The current system with Cargo feature was an experiment to see if it was a good idea.

It turns out that:

 - It is not very intuitive.
 - I have never seen someone using it (including myself, I don't use it).
 - It puts the code that checks whether an action is possible and the code that does the action in two different places.
 - Some things are very difficult/annoying to handle, like vertex formats.
 - The checks at initialization could as well be done by the user's code.
 - It brings no performance improvements anyway.

I think it should just be replaced by `NotSupported` errors.

tomaka:
Close #1073


cmr:
:+1:

PR #1097, opened 2015-07-15
Add an alternate set_finish() method to Frame

TyOverby:
When a struct contains a `Frame`, it is really hard to call `finish()` on the frame contained within.  

This patch  adds a new method called `set_finish()` which does all the cleanup that `finish()` does, with the only difference being that it takes the `Frame` by `&mut self` instead.  This makes it easier to trigger when it is contained in another type.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -911,6 +911,10 @@ pub enum SwapBuffersError {
     /// sleep and wakes it up later. However any OpenGL implementation can theoretically lose the
     /// context at any time. Can only happen if calling `is_context_loss_possible()` returns true.
     ContextLost,
+    /// The buffers have already been swapped.
+    ///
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -911,13 +911,14 @@ pub enum SwapBuffersError {
     /// sleep and wakes it up later. However any OpenGL implementation can theoretically lose the
     /// context at any time. Can only happen if calling `is_context_loss_possible()` returns true.
     ContextLost,
+    AlreadySwapped,
```
tomaka: Documentation is missing

tomaka:
I'll merge after you add docs.

TyOverby:
Comments added and tests pass.  I'll merge this if you don't mind.

PR #1103, opened 2015-07-17
Add all constructors to buffers

### LINKED ISSUE #882: Be more specific in buffers creation
I feel that the `dynamic()` method creating persistent mapped buffers is too high-level/confusing.

Instead we should add `persistent_if_supported`, `persistent`, and `persistent_or_dynamic`.

tomaka:
Close #882


tomaka:
Forgot about the mapping-related flags!

PR #1106, opened 2015-07-19
Fix the call to StencilFuncSeparate for gl::FRONT.

fjolliton:
While the state was tracked correctly, the settings for gl::BACK
were applied instead.


tomaka:
Thanks!

PR #1154, opened 2015-08-10
[Acceptable breaking change] Reform the To*Attachment traits

tomaka:


tomaka:
circle-ci fails to download the dependencies, ignoring it.

Issue #1158, opened 2015-08-10
Support ARB_ES3_2_compatibility


tomaka:
https://www.opengl.org/registry/specs/ARB/ES3_2_compatibility.txt

tomaka:
And OES_primitive_bounding_box as well

PR #1166, opened 2015-08-10
[Acceptable breaking change] Add support for specifying the primitive bounding box

tomaka:
Depends on https://github.com/bjz/gl-rs/pull/351

cc #1158 

`DrawParameters` no longer implements `Copy` (since `Range` doesn't implement `Copy`), so this is a small breaking change.


src/draw_parameters/mod.rs
```diff
@@ -826,6 +826,21 @@ pub struct DrawParameters<'a> {
     /// potentially trigger a `ProvokingVertexNotSupported` error. Most notably OpenGL ES doesn't
     /// support anything else but `LastVertex`.
     pub provoking_vertex: ProvokingVertex,
+
+    /// Hint for the GPU of the bounding box of the geometry.
+    ///
+    /// If you're using geometry shaders or tessellation shaders, it can be extremely advantageous
+    /// for the GPU to know where on the screen the the primitive. This field specificies the
```
meh: `the the` primitive.

tomaka:
Thanks

PR #1180, opened 2015-08-24
Support signed and unsigned int uniform vector types

msiglreith:
Note: The context versions for the `gl.Uniform{1|2|3|4}ui{v}` should probably be changed. [OpenGL](https://www.opengl.org/sdk/docs/man/html/glUniform.xhtml) and [GL ES](https://www.khronos.org/opengles/sdk/docs/man3/html/glUniform.xhtml) support these since 3.0 according to the docs. But I couldn't test it at the moment and used [these] (https://github.com/tomaka/glium/blob/master/src/program/uniforms_storage.rs#L79-L89) versions insteads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
  - Renamed `BufferView` to `Buffer`. `BufferView` still exists for backward compatibility.
  - Add support for the `GL_OES_element_index_uint` extension.
  - Fixed OpenGL ES 3.2 not working.
+ - Add support for `IntVec{2|3|4}` and `UnsignedIntVec{2|3|4}` uniform types.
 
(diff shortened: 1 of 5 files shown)


tomaka:
Thanks!

The GLSL compiler detects for us whether the uniform types are supported.
If a shader that contains a `ivec2` uniform compiles just fine, then glium automatically assumes that `glUniform2iv` is supported as well (it wouldn't make sense if some uniform types were supported but no way to set their value).

The version checks are only here to handle the `GL_ARB_shader_objects` extension.

msiglreith:
Ah, I wasn't aware of this.
Added a line to the `CHANGELOG.md`.

PR #1182, opened 2015-08-24
Add a work-around for Radeon drivers crashing with 32+ texture units

### LINKED ISSUE #1181: Windows/AMD driver can crash when using texture unit >= 32
There appears to be a bug with AMD's drivers which can cause crashes when texture unit large enough is used in rendering.

Sample project demonstrating the issue: https://github.com/neivv/glium-amd-crash
It crashes within seconds on Windows 8.1 64-bit with Radeon HD 7500G using Catalyst 15.200.1062.1004 (The project is built as 32-bit .exe, I haven't tested 64-bit)

After doing some reverse engineering of the atioglxx.dll, there was an array of [32 * 6] entries for texture units, which was overflown corrupting nearby memory locations and crashing the driver.
As glium uses every texture unit slot once before reusing any of the unused ones, even a simple program is able to trigger the crash.

Possibly relevant issue of a game crashing with AMD OpenGL: https://steamcommunity.com/app/244160/discussions/1/617328967244398617/
Using the older driver .dll from the link, the sample program works and the older .dll reports 32 as GL_MAX_COMBINED_TEXTURE_IMAGE_UNITS.

tomaka:
Fix #1181 

cc @neivv
This should fix it I guess.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  - Add support for the `GL_OES_element_index_uint` extension.
  - Fixed OpenGL ES 3.2 not working.
  - Add support for `IntVec{2|3|4}` and `UnsignedIntVec{2|3|4}` uniform types.
+ - Add a work-around for Radeon drivers crashing with 32+ texture units.
 
 ## Version 0.8.5 (2015-08-12)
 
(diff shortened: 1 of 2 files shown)


neivv:
Yep, this fixes the issue on my system.

PR #1184, opened 2015-08-25
Fix compiler warnings on nightly.

nicokoch:
This commit fixes some of the compiler warnings, which have been
introduced with recent changes on the nightly channel.
I made sure this did not influence building on stable (thus we're still
using str::connect()).

Not really sure if this is a breaking change, so I decided to create a seperate PR for this.

### DIFF
--- a/src/texture/mod.rs
+++ b/src/texture/mod.rs
@@ -110,7 +110,7 @@ include!(concat!(env!("OUT_DIR"), "/textures.rs"));
 
 /// Represents a layer of a cubemap.
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
-#[allow(missing_docs)]      // TODO: 
+#[allow(missing_docs)]      // TODO:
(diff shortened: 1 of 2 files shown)


src/texture/mod.rs
```diff
@@ -505,9 +505,9 @@ pub trait Texture3dDataSource<'a> {
 ///
 /// You are especially encouraged to implement this trait with the parameter `(u8, u8, u8, u8)`,
 /// as this is the only format that is guaranteed to be supported by OpenGL when reading pixels.
-pub trait Texture3dDataSink<T> {
+pub trait Texture3dDataSink<T: ToOwned> {
```
tomaka: This one isn't needed.

tomaka:
r+ after the change, thanks!

nicokoch:
Done!

tomaka:
circle-ci not working again :-/

PR #1191, opened 2015-08-26
Add support for double precision uniforms.

nicokoch:
These were missing in glium until now. Require `GL_ARB_gpu_shader_fp64` .

I'm not really sure if I did the OpenGL extension handling right, so maybe someone should look over that. Only compile tested for now because the Mesa drivers for my card only support OpenGL 3.3.


tomaka:
Thanks

PR #1192, opened 2015-08-26
Add range of tests for simple uniform types

nicokoch:
These (kinda) analogous to the tests in attr.rs .
If a `#version 400` shader is supplied by someone (my graphics card is incompatible), the commented out double tests can be used to test #1191 .

### DIFF
--- a/tests/uniforms.rs
+++ b/tests/uniforms.rs
@@ -171,3 +171,101 @@ fn uniform_wrong_type() {
 
     display.assert_no_error(None);
 }
+
+macro_rules! uniform_test(
+    ($name:ident, $glsl_ty:expr, $value:expr) => (
+        #[test]
+        fn $name() {
+            let display = support::build_display();
+            let (vb, ib) = support::build_rectangle_vb_ib(&display);
+            // TODO Supply #version 400 for doubles and #version 100 for fallback (but uint needs 130 I believe)
+            let program = glium::Program::from_source(&display,
+                &format!("
+                    #version 130
+                    attribute vec2 position;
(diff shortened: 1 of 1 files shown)


tomaka:
Ah, good idea.

In my opinion the tests should try creating a program with `#version 400`, then if it doesn't work with `#version 140` (the smallest version supported by core OpenGL), then `#version 110` and then `#version 100`.

nicokoch:
Thanks for your suggestions. As I'm not really that familiar with GLSL, would you mind providing a `#version 400` shader? The rest I will probably be able to do on my own.

tomaka:
With versions 400 and 140 you have to replace `attribute` with `in`, `varying` in vertex shaders with `out`, and `varying` in fragment shaders with `in`. The rest is the same.

nicokoch:
Ah nice, that helps a lot.

Is the `program!()` macro appropriate for this? For me it panics with `no version found` here, but maybe I'm doing something wrong.

tomaka:
> For me it panics with no version found here, but maybe I'm doing something wrong.

Yeah, it panicks, that's why it's not appropriate.

nicokoch:
Ok, nice. One last problem: How would I check if `double` or `uint` are supported in this context, and how would I deal with it if it's not? (Probably just return early?)

tomaka:
If they are not supported then none of the shaders will compile.

tomaka:
Thanks

nicokoch:
Since #1191 has been merged, the tests in lines 296-313 (currently commented out) can be activated. Want me to create a seperate PR for that or do it youself?

PR #1199, opened 2015-08-29
Check for maximum number of possible color attachments to a framebuffer

msiglreith:


tomaka:
- OpenGL ES 2 doesn't support `GL_MAX_COLOR_ATTACHMENTS` (the value should be 1) unless this extension is present: https://www.khronos.org/registry/gles/extensions/NV/NV_fbo_color_attachments.txt

 - The `validate_layered` and `validate_regular` functions should also make this check and return an error, in order to avoid panicks. Note that for the moment it would always panic, but I plan to change that in the future, so the `validate_*` functions should be adapted.

 - A test with `#[should_fail]` should be added for this. It should create, say, 32 textures and try to attach them all.

msiglreith:
Thanks for the feedback!
I tried to address all the points you mentioned. The test case might be written a bit nicer, but I don't know how to do it better atm.

(cc #560)

src/context/capabilities.rs
```diff
@@ -389,6 +392,21 @@ pub unsafe fn get_capabilities(gl: &gl::Gl, version: &Version, extensions: &Exte
...
+               extensions.gl_arb_framebuffer_object || extensions.gl_ext_framebuffer_object ||
+               extensions.gl_nv_fbo_color_attachments
+            {
+                let mut val = 4;
+                gl.GetIntegerv(gl::MAX_COLOR_ATTACHMENTS, &mut val);
+                val
+            } else if version >= &Version(Api::GlEs, 2, 0) {
+                1
+            } else {
+                0
```
tomaka: Should be `unreachable!()`

tomaka:
Oh I forgot there was already a `TooManyColorAttachments` error. Please use it instead of `ColorAttachmentBindPointNotSupported`.

tomaka:
Thanks

PR #1219, opened 2015-09-08
update macros to use fully qualified result type.

cpjreynolds:
The `implement_buffer_content` and `implement_uniform_block` macros did not use a fully qualified result type, which could cause interference with custom result types used in the same module. (Just ran into this issue myself).


tomaka:
Thanks!

PR #1220, opened 2015-09-08
Add possibility to clear with an sRGB color

tomaka:
cc @eddyb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
  - The blending function now uses a dedicated struct named `Blend`.
  - Glium no longer panics in case of OpenGL error. It prints a message with backtrace on stdout instead.
  - Removed the `DrawParametersBuilder` struct.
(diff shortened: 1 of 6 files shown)


Yamakaky:
Did you consider instead to introduce an enum Color with RGB and SRGB variants ?

tomaka:
There are other functions that will need a dinstinction between RGB and sRGB, for example reading from the color buffer or reading/writing to a texture. These functions will take as parameter or return a `Vec<(f32, f32, f32)>`.

I think it would be too annoying to manipulate a `Vec<Srgb>` or a `SrgbVec<(f32, f32, f32, 32)>` or something similar, so I think it's better to name them `read_rgb` and `read_srgb` for example.

Therefore for consistency I wrote the `clear_color` API to be similar to these.

Yamakaky:
With implementations for From<SRGB> and From<(f32, f32, f32, f32)>, it would be easier to manipulate.

PR #1245, opened 2015-09-16
Add per_instance for VertexBufferSlice

tomaka:
Also optimizes the "instancing" example.


tomaka:
The optimisation was here because someone reported that this example was slow.

It turns out that this change doesn't improve the situation, so I'm going to revert the optimisation.

tomaka:
Closing for #1260

Issue #1244, opened 2015-09-15
The triangle_* tests fail on  my system


nicokoch:
```
cargo test triangle
[...]skipping some bloat lines[...]

Running target/debug/indices-939bdd9e70e3f0a4

running 6 tests
test triangle_fan_noindices ... FAILED
test triangle_strip ... FAILED
test triangle_strip_noindices ... FAILED
test triangle_fan ... FAILED
test triangles_list_noindices ... FAILED
test triangles_list ... FAILED

failures:

---- triangle_fan_noindices stdout ----
        thread 'triangle_fan_noindices' panicked at 'assertion failed: `(left == right)` (left: `(0, 0, 0, 0)`, right: `(255, 0, 0, 255)`)', tests/indices.rs:266


---- triangle_strip stdout ----
        thread 'triangle_strip' panicked at 'assertion failed: `(left == right)` (left: `(0, 0, 0, 0)`, right: `(255, 0, 0, 255)`)', tests/indices.rs:103


---- triangle_strip_noindices stdout ----
        thread 'triangle_strip_noindices' panicked at 'assertion failed: `(left == right)` (left: `(0, 0, 0, 0)`, right: `(255, 0, 0, 255)`)', tests/indices.rs:238


---- triangle_fan stdout ----
        thread 'triangle_fan' panicked at 'assertion failed: `(left == right)` (left: `(0, 0, 0, 0)`, right: `(255, 0, 0, 255)`)', tests/indices.rs:130


---- triangles_list_noindices stdout ----
        thread 'triangles_list_noindices' panicked at 'assertion failed: `(left == right)` (left: `(0, 0, 0, 0)`, right: `(255, 0, 0, 255)`)', tests/indices.rs:212


---- triangles_list stdout ----
       thread 'triangles_list' panicked at 'assertion failed: `(left == right)` (left: `(0, 0, 0, 0)`, right: `(255, 0, 0, 255)`)', tests/indices.rs:77



failures:
    triangle_fan
    triangle_fan_noindices
    triangle_strip
    triangle_strip_noindices
    triangles_list
    triangles_list_noindices

test result: FAILED. 0 passed; 6 failed; 0 ignored; 0 measured

thread '<main>' panicked at 'Some tests failed', ../src/libtest/lib.rs:255
```
I noticed this isn't happening on travis-ci, so it actually might be a driver bug. I'm using the i965 driver. My OpenGL system info:

```
glxinfo | grep -i opengl  
[...]

Issue #310, opened 2015-01-17
Rewrite tests to write and read on a texture instead of the default framebuffer


tomaka:
The implementation is free to discard some pixels when written to the default framebuffer if the window or part of the window is not visible. This is bad for tests.

tomaka:
Search for `read_front_buffer` to see what's left to be converted.

PR #1249, opened 2015-09-16
Fix triangle tests to read from a texture

nicokoch:
as discussed in #310 and #1244 .
Also fix some compiler warnings.


tomaka:
Thanks!

Do the tests work with this change?

nicokoch:
Yep

Issue #1157, opened 2015-08-10
Support GL_ARB_gpu_shader_int64


tomaka:
https://www.opengl.org/registry/specs/ARB/gpu_shader_int64.txt

nicokoch:
I have implemented this already, but it depends on bjz/gl-rs#352 to be merged an published, so I'm gonna wait with the PR

tomaka:
There's a little problem: updating to a latest version of gl_generator triggers a stack overflow in rustc when compiling glium.

nicokoch:
Oh that sucks. Is there any solution on the way?

tomaka:
No :-/

nicokoch:
If I check out the git version of GL_generator and compile glium with that version it works just fine. I am on linux. Are you sure this problem still occurs?

tomaka:
I updated #1011, we'll see what travis says.

tomaka:
gl_generator 0.0.28 has been published

PR #1251, opened 2015-09-17
Implement i64 and u64 uniform types

nicokoch:
cc #1157 
I'm not really sure if additional work is required to support 64 Bit integer attributes.
Also you should probably run the tests locally once to make sure it really works, since my drivers do not support this feature yet and thus the tests return early.

### DIFF
--- a/src/program/uniforms_storage.rs
+++ b/src/program/uniforms_storage.rs
@@ -75,7 +75,7 @@ impl UniformsStorage {
                     if $ctxt.extensions.gl_arb_gpu_shader_int64 {
                         $ctxt.gl.$uniform($($params),+)
                     } else {
-                        panic!("Double precision integers are not supported on this system.")
+                        panic!("64 bit integers are not supported on this system.")
                     }
                 }
             )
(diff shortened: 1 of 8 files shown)


src/program/uniforms_storage.rs
```diff
@@ -57,13 +57,25 @@ impl UniformsStorage {
...
+            )
+        );
+
+        macro_rules! uniform_i64(
+            ($ctxt:expr, $uniform:ident, $($params:expr),+) => (
+                unsafe {
+                    if $ctxt.extensions.gl_arb_gpu_shader_int64 {
+                        $ctxt.gl.$uniform($($params),+)
+                    } else {
+                        panic!("Double precision integers are not supported on this system.")
```
tomaka: `Double precision integers` doesn't really make sense.

tomaka:
Arg, my local rustc triggers a stack overflow.

tomaka:
My driver doesn't support `GL_ARB_gpu_shader_int64`.

tomaka:
I fixed the stack overflow in #1252

nicokoch:
Very nice! That is a dirty workaround though :dancers:

PR #1260, opened 2015-09-21
VertexBufferSlice instancing

acolley:
This adds the `per_instance()` method to the `VertexBufferSlice` struct as I had a need need for this in my current application. Also added an example that demonstrates this feature that is based off the existing instancing example.


tomaka:
Thanks!
However I don't think it's worth an example. The example is too similar to instancing.

I'll merge after you remove the example and rebase the PR.

acolley:
Just noticed you made an almost identical update several days before I opened this :)

tomaka:
I don't know what you did with your commits :o
I'll merge anyway. Thanks.

PR #1273, opened 2015-10-03
Fix for multiple objects in wavefront .obj file.

gfokkema:
Fixes the loader used in the teapot example.
Might spare someone else a few hours ;)


tomaka:
Thanks

PR #1309, opened 2015-10-19
Fix several compiler warnings

nicokoch:


tomaka:
Thanks, however I'd prefer not to remove the `unwrap()`. They are real errors that could happen and that should be handled instead of being ignored.

nicokoch:
You mean instead of `unwrap` we should match the result right?

tomaka:
Just leave the warning for now. The error should be propagated.

nicokoch:
I just updated this PR.

I also tried running [clippy](https://github.com/Manishearth/rust-clippy) on this crate, and ran into a compiler [bug](https://github.com/rust-lang/rust/issues/29174).

Issue #1246, opened 2015-09-16
Rename ClockWise and CounterClockWise to Clockwise and CounterClockwise


tomaka:

tomaka:
Fixed in #1288

PR #1312, opened 2015-10-21
Update tutorial for ClockWise -> Clockwise rename.

wheals:


wheals:
Also, the commit that renamed them didn't actually close #1246.

tomaka:
Thanks!

PR #1320, opened 2015-10-24
Add DefaultFramebuffer and DefaultFramebufferAttachment

tomaka:
Mostly useless for now.


tomaka:
More precisions:

It's currently impossible to read/write anything else than the back left buffer.

This code is a draft for what the API will look like in the future, but for the moment it only allows access to the back left buffer as well. The reason is that the rest of glium needs a lot of changes to access another buffer.

And for these other changes, the `DefaultFramebufferAttachment` enum will be used.

Issue #1314, opened 2015-10-22
VertexBuffer::slice doesn't accept RangeFrom


Yamakaky:
It only accepts Range.

Yamakaky:
https://doc.rust-lang.org/collections/range/trait.RangeArgument.html would be perfect if it wasn't unstable.

Esption:
If/when [RangeArgument](https://doc.rust-lang.org/collections/range/trait.RangeArgument.html) becomes stable it could be used for this it looks like, but until then it's probably not going to be possible without re-implementing it or making separate functions.

@tomaka would you be fine with a temporary re-implementation of RangeArgument while waiting for it to be become stable?  If so I can get a patch for this done then.  I'm not really sure how else this could be done otherwise.

Esption:
@Yamakaky I don't really have some code to test this on atm, could you tell me if that works? https://github.com/Esption/glium

PR #1321, opened 2015-10-24
Change Range to RangeArgument

Esption:
#1314

In order to avoid needing to use the unstable RangeArgument I just rewrote.  All we should need to do if/when it becomes stable is change to use glium::buffer::RangeArgument to collections::range::RangeArgument and *also* will probably be a breaking change because users will need to change that too.  Granted, we could just deprecate the glium one for a while for people to switch over, I guess.

This is also a breaking change now, because people will need to import RangeArgument to use buffer slice.

### DIFF
--- a/src/buffer/mod.rs
+++ b/src/buffer/mod.rs
@@ -67,9 +67,6 @@ use std::fmt;
 use std::mem;
 use std::slice;
 
-//For temp RangeArgument
-use std::ops::{Range, RangeTo, RangeFrom, RangeFull};
-
 mod alloc;
 mod fences;
 mod view;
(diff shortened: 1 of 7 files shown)


tomaka:
Why would you need to import the trait?

src/buffer/mod.rs
```diff
@@ -316,3 +319,44 @@ impl BufferType {
         }
     }
 }
+
+/// Temporary re-implemenation of RangeArgument from stdlib while
+/// waiting for it to become stable
+pub trait RangeArgument<T> {
```
tomaka: Could you move the definition in `src/utils.rs` instead?

src/buffer/view.rs
```diff
@@ -658,15 +658,16 @@ impl<'a, T> BufferSlice<'a, [T]> where [T]: Content + 'a {
     /// This method builds an object that represents a slice of the buffer. No actual operation
     /// OpenGL is performed.
     #[inline]
-    pub fn slice(&self, range: Range<usize>) -> Option<BufferSlice<'a, [T]>> {
-        if range.start > self.len() || range.end > self.len() {
+    pub fn slice<R: RangeArgument<usize>>(&self, range: R) -> Option<BufferSlice<'a, [T]>> {
+		let clos = |val: &usize| { *val };
```
tomaka: Wrong indentation (below too)

src/buffer/view.rs
```diff
@@ -658,15 +658,16 @@ impl<'a, T> BufferSlice<'a, [T]> where [T]: Content + 'a {
...
+        if range.start().map_or(0, &clos) > self.len() || range.end().map_or(0, &clos) > self.len() {
             return None;
         }
 
         Some(BufferSlice {
             alloc: self.alloc,
-            bytes_start: self.bytes_start + range.start * mem::size_of::<T>(),
-            bytes_end: self.bytes_start + range.end * mem::size_of::<T>(),
+            bytes_start: self.bytes_start + range.start().map_or(0, &clos) * mem::size_of::<T>(),
```
tomaka: Anyway, I'd prefer if you just repeated the closure instead of defining it at the top.
Just something like `.map_or(0, |&v| v)`.

Yamakaky:
Btw, do you know when the upstream trait will be stable? Because 1.4 will be out in a [few days](https://github.com/rust-lang/rust/milestones)

Esption:
Don't you need to import traits to use them? Maybe I'm crazy? I dunno.

tomaka:
You would need to import the trait if you called `start()` and `end()` manually. You don't need to import it just to pass a parameter to a function that requests the trait.

Esption:
Alright, so I am crazy lol.  So anyway, not a breaking change then.

tomaka:
Thanks!

PR #1323, opened 2015-10-25
Update genmesh

tomaka:
Fixes the warnings about the deprecated `Sized` stuff.


tomaka:
Grmlblb, `genmesh` has a `*` dependency.

Issue #1284, opened 2015-10-09
SrgbTexture2d has no write() method


HeroesGrave:
A quick look through the docs shows that this is the case for a few other texture types too.

PR #1326, opened 2015-10-26
Fix missing write() on some texture types.

Esption:
#1284

While this is being touched, before merging this is there any other `TextureType` that should have `write()`? Currently it's just `Regular` and `Srgb` along with the Compressed equivalents of those two.  I'm not really sure about `Depth`/`Stencil`/etc should have them.

Also, this fixes some screw-ups from my last PR (I use tabs personally, I finally got `git config` set up correctly so it will auto-convert tabs/spaces for me... sorry about that.)


tomaka:
Yes, other texture types should eventually have `write`, but with a different API.
I still have to figure out the design of this.

Thanks, I'll merge when CI passes.

PR #1328, opened 2015-10-28
Add trait implementations for Sampler

SpaceManiac:
Sampler doesn't currently implement any of the traits it reasonably could. Copy and Clone are not derived because they would place incorrect bounds on T due to the reference field.


tomaka:
Thanks!

PR #1331, opened 2015-11-04
0.11.0

tomaka:


bvssvni:
Yay! I have been waiting to update glium_graphics for a while now. Using different major version on the image library causes some problems with integration.

Issue #1072, opened 2015-07-08
Support externally managed textures


ryanstew:
I started playing with this, but opening this issue for comment/discussion. Maybe this is too ugly and out of scope for glium.

The specific use case is the Oculus Compositor. The Oculus SDK has an API for creating one or more texture sets that you are expected to draw to each frame. You then submit the frame to the SDK, and it handles composition, distortion, and presentation to the headset from the drawn texture. Because of this handshake, the Oculus SDK creates and owns the OpenGL texture handle.

To present to the compositor via glium, the moral equivalent of binding the Oculus-owned texture to a `SimpleFrameBuffer` is required. I don't see a way to do that currently.

One approach is to add something like `unsafe fn from_unowned()` to `TextureAny` and a similar method to all the texture types. This would wrap an externally managed texture id for broad use around glium.

Making it safe is challenging (impossible?) since lifetime/ownership is of the handle must be shared. I suppose there are more localized changes (e.g. extending `ColorAttachment`/`DepthAttachment` with an external texture reference type) that would limit the unsafe scope.

tomaka:
The problem is that there is a `HashMap<FramebufferAttachments, FramebufferObject>` (pseudo code) in the context.

When you draw on a texture for the first time, it creates a framebuffer object and adds it to the hashmap. In the destructor of `TextureAny`, all framebuffer objects where one of the attachments is the given texture are destroyed.

If you allow drawing on a texture by id, glium would need to create a temporary framebuffer object which would be recreated every time. The alternative is to make the user manually handle the cleanup of the framebuffer objects when a texture is destroyed, but that's not nice either. If the framebuffer is not destroyed, glium could reuse an existing framebuffer object with a newly-created texture.

The alternative to add `from_unowned` is nicer. Adding `into_inner` (or a similar name), which cleans up framebuffer objects but doesn't destroy the texture, could be useful as well.

tomaka:
Or maybe add `from_id(id: u32, owned: bool)`.
If `owned` is true, the texture will be destroyed by glium.

PR #1344, opened 2015-11-17
TextureAny::from_id

ryanstew:
Add from_id to texture types to allow them to point to an externally-allocated texture. See #1072 for details.

### DIFF
--- a/src/texture/any.rs
+++ b/src/texture/any.rs
@@ -69,6 +69,53 @@ pub struct TextureAny {
     levels: u32,
     /// Is automatic mipmap generation allowed for this texture?
     generate_mipmaps: bool,
+
+    /// Is this texture owned by us? If not, we won't clean it up on drop.
+    owned: bool
+}
+
+fn extract_dimensions(ty: Dimensions)
+                      -> (u32, Option<u32>, Option<u32>, Option<u32>, Option<u32>)
+{
+    match ty {
+        Dimensions::Texture1d { width } => (width, None, None, None, None),
+        Dimensions::Texture1dArray { width, array_size } => (width, None, None, Some(array_size), None),
+        Dimensions::Texture2d { width, height } => (width, Some(height), None, None, None),
+        Dimensions::Texture2dArray { width, height, array_size } => (width, Some(height), None, Some(array_size), None),
+        Dimensions::Texture2dMultisample { width, height, samples } => (width, Some(height), None, None, Some(samples)),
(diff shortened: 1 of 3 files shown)


src/texture/any.rs
```diff
@@ -69,6 +69,53 @@ pub struct TextureAny {
...
+        Dimensions::Texture2dArray { .. } => gl::TEXTURE_2D_ARRAY,
+        Dimensions::Texture2dMultisample { .. } => gl::TEXTURE_2D_MULTISAMPLE,
+        Dimensions::Texture2dMultisampleArray { .. } => gl::TEXTURE_2D_MULTISAMPLE_ARRAY,
+        Dimensions::Texture3d { .. } => gl::TEXTURE_3D,
+        Dimensions::Cubemap { .. } => gl::TEXTURE_CUBE_MAP,
+        Dimensions::CubemapArray { .. } => gl::TEXTURE_CUBE_MAP_ARRAY,
+    }
+}
+
+unsafe fn generate_mipmaps(ctxt: &CommandContext,
```
ryanstew: Open question: where mipmaps are regenerated on upload, this logic is slightly different (the GLES check is not in place). Is that a bug, or intentional?

tests/texture_draw.rs
```diff
@@ -34,30 +30,81 @@ macro_rules! texture_draw_test {
...
+            let (vb, ib) = support::build_rectangle_vb_ib(&display);
+
+            let program = create_program!(display, $glsl_ty, $glsl_value);
+
+            let empty_texture = glium::texture::$tex_ty::empty_with_format(&display,
+                                                                           glium::texture::$format_ty::$format_value,
+                                                                           glium::texture::MipmapsOption::NoMipmap,
+                                                                           $($dims),+).unwrap();
+            let texture = unsafe {
+                glium::texture::$tex_ty::from_id(&display,
```
ryanstew: I assume it's safe to just repeat the format here and trust that Glium didn't change the underlying format, as I can't find an easy way to convert the internal format of a texture back out into an UncompressedFloatFormat.

src/texture/any.rs
tomaka: That would be a bug.

matklad:
May I suggest changing the type of `owned: bool` from `bool` to two element `enum` in the sake of readability? 

At the call site, `Owner::Callee` would be more self documenting then `true`.

tomaka:
Let's merge this.

The code is not very clean, but the textures code in general is really dirty.

PR #1347, opened 2015-11-21
Update image dependency

nstoddard:
This seems to be necessary to work around rust-lang/cargo#2132


tomaka:
Will be fixed by https://github.com/tomaka/glium/pull/1337 instead

PR #1357, opened 2015-12-02
Bump cgmath version to 0.5

bitshifter:
Update cgmath to latest.

I was using cgmath 0.5 in my project, which clashed with the cgmath 0.4 being used in glium.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ features = []
 optional = true
 
 [dependencies.cgmath]
-version = "0.4"
+version = "0.5"
 optional = true
 
 [dependencies.nalgebra]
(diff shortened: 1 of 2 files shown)


bjz:
Thanks!

We switched a number of functions over from pass-by-ref to pass-by-value. You'll need to fix [these usages here](https://travis-ci.org/tomaka/glium/jobs/94303777#L510-L559). Be sure to compile the tests with:

```sh
cargo test --features 'cgmath'
```

bitshifter:
Ah, I thought that seemed far too easy :)

I didn't notice cgmath wasn't a default feature. I'll take a look.

bjz:
\o/

tomaka:
I think I'm going to remove support for cgmath and nalgebra, like I did for image.
The From/Into traits should be good enough I think.

bjz:
:+1:

bitshifter:
Seems reasonable, should I just close this PR then?

PR #1362, opened 2015-12-03
Remove support for cgmath and nalgebra

### LINKED ISSUE #364: remove cgmath support
cgmath is deprecated: https://github.com/bjz/cgmath-rs/issues/144

Removing support in favor of nalgebra will lead to a less confusing API.

tomaka:
Close #1357 
Close #364

Will avoid problems in the future with incompatibilities between glium and math libraries.

I'm not totally confident about how usable it is however.
Writing `Into::<[[f32; 4]; 4]>::into(matrix)` is a bit cumbersome, and not that great since you expose the fact that the matrix has to be column-major.


bitshifter:
Maybe a separate binding library could help?

MasonRemaley:
The Features section of README.md still references `nalgebra` and `cgmath` from before they were removed.

tomaka:
@MasonRemaley Fixed by #1424

tyoc213:
Well I think also http://tomaka.github.io/glium/glium/uniforms/trait.AsUniformValue.html  there are links to nalgebra and cgmath?

Also it would be nice to have some examples... well (still a rust noob) I tried different ways... still not much confortable with rust type system and messages I guess...

```
         let m:Basis3<f32> =  Rotation3::from_axis_angle(Vector3::new(0.0f32, 0.0, 1.0).normalize(), deg(180.0f32).into());
         let m = m.as_ref();
         let m4 = Matrix4::new(m[0][0], m[0][1], m[0][2], f32::zero(),
                      m[1][0], m[1][1], m[1][2], f32::zero(),
                      m[2][0], m[2][1], m[2][2], f32::zero(),
                      f32::zero(), f32::zero(), f32::zero(), f32::one());
         let thingUniform = Into::<[[f32; 4]; 4]>::into(m4);
```
I was able to "extract" that code from https://github.com/bjz/cgmath/blob/6cf7831275de461d450ef514a2148b835c59a01f/src/matrix.rs#L1060 but for some reason I can't use it the way "I imagine"... ```cgmath::Matrix4::from(m.as_ref())``` or some like that.

bjz:
Currently I have some utility functions in my game project to help the type inference:

```rust
pub fn array_v3(v: Vector3<f32>) -> [f32; 3] {
    v.into()
}

pub fn array_p3(p: Point3<f32>) -> [f32; 3] {
    p.into()
}

pub fn array_m4(m: Matrix4<f32>) -> [[f32; 4]; 4] {
    m.into()
}
```

I use them like so:

```rust
&uniform! {
    color:      color,
    light_dir:  math::array_v3(light_dir),
    model:      math::array_m4(Matrix4::identity()),
    view:       math::array_m4(self.camera.view),
    proj:       math::array_m4(self.camera.projection),
    eye:        math::array_p3(self.camera.position),
},
```

Perhaps it might make sense for me to include those conversion functions in a submodule of cgmath?

bjz:
I've made a PR adding these similar functions to cgmath: bjz/cgmath#309

PR #1365, opened 2015-12-05
Fix spelling in book

awjwdev:


tomaka:
Thanks!
Actually that should be `process`. The `u` is not a typo, it's because `processus` is the translation of `process` in my mother tongue.

awjwdev:
@tomaka My apologies! Force pushed using `process` instead.

PR #1370, opened 2015-12-11
Update BlendColor when the blending equation uses ConstantAlpha

simias:
The current code only checks for ConstantColor so the blend constant doesn't get updated if you try to use `ConstantAlpha` or `OneMinusConstantAlpha`


tomaka:
Thanks!

PR #1371, opened 2015-12-15
[WIP] Buffers rework

### LINKED ISSUE #1325: Idea: split buffer types into DynamicBuffer, ImmutableBuffer, PersistentBuffer
...with different APIs.

Note that by "dynamic" I mean "handled by OpenGL".

The way you handle a non-persistent buffer is really different from the way you use a dynamic buffer, especially with the triple buffering idiom.

This split would allow to:

- Change the size of a dynamic buffer at runtime.
- Forbid all modifications of an immutable buffer.
- Use `glMapBuffer(UNSYNCHRONIZED)` to map a persistent buffer if persistent buffers are not supported.

tomaka:
Close #1325 

- [ ] Make `VerticesSource` and `IndicesSource` private ([breaking-change]).
- [x] Splits `Buffer` between `DynamicBuffer`, `ImmutableBuffer` and `PersistentBuffer`.
- [ ] Split the underlying implementations in order to produce more optimized calls.
- [ ] Make the fencing strategy of persistent buffers explicit and that can be chosen by the user.
- [ ] Rework index buffers.
- [ ] Rework vertex buffers.
- [ ] Rework buffer textures.
- [ ] Rework multidraw indirect buffers.
- [ ] Rework uniform buffers.
- [ ] Rework pixel buffers.


nicokoch:
Are you still working on this?

tomaka:
No :-/

As I said I'm thinking of reorienting glium towards a more OpenGL 2-OpenGL 3-based experience.
For example I've been thinking of removing persistant-mapped buffers entirely.

PR #1383, opened 2015-12-29
Fixing Issue #1087 

Roxxik:
Replace the OpenGL logo with a Rusty logo #1087


tomaka:
Thanks for the PR!

But I'm a bit puzzled about this, because the Rust logo has a transparent background. Someone who reads the tutorial will probably wonder why the background appears black.
On the other hand, this would be a good opportunity to modify the tutorial to talk about blending and make it transparent.

Roxxik:
Yeah when using the screenshot functionality (copy framebuffer to imagebuffer and saving it) i'll get a transparent background, but if i use the screenshot functionality from gimp it'is black.

Issue #741, opened 2015-04-13
Switch to `Into`, `From`, etc. traits


tomaka:
This library has several traits named `IntoFoo`. They should be replaced by `Into<Foo>`.

PR #1384, opened 2015-12-29
turned IntoVerticesSource into Into<VerticesSource> Issue #741

Roxxik:
fixing Issue #741 , there are just some IntoIterator's and IntoIter's left

### DIFF
--- a/src/vertex/buffer.rs
+++ b/src/vertex/buffer.rs
@@ -4,7 +4,7 @@ use std::ops::{Deref, DerefMut};
 use utils::range::RangeArgument;
 
 use buffer::{Buffer, BufferSlice, BufferMutSlice, BufferAny, BufferType, BufferMode, BufferCreationError, Content};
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks!

This is a breaking change though, so I'll merge it before 0.14.

Roxxik:
No problem ;) i'm just getting into OSS, git and rust and am just picking some low-hanging fruits/fixes

daboross:
Would it make sense to merge this for 0.20.0? It seems like a good change and there is already a breaking change in the repository right now.

I've rebased the changes onto master and fixed the one new merge conflict in https://github.com/daboross/glium/tree/i741 - I don't know if it would be better to PR this to roxxik/glium:i741 or make a new PR to glium/glium though.

tomaka:
Yeah a new PR would be better

daboross:
Alright- opened #1661.

PR #1389, opened 2016-01-02
Use identity() instead of from_scale(1.0)

IvanUkhov:
Hi,

In #1380, I used `Matrix4::from_scale(1.0)` as a substitute for `Matrix4::one()`. @bjz, however, [pointed out](https://github.com/tomaka/glium/pull/1380/files#r48675032) that a better way would be to use `Matrix4::identity()`.

Regards,
Ivan


tomaka:
Thanks

PR #1390, opened 2016-01-02
FPS cam

Roxxik:
Code is mostly taken from tutorial-12.rs with added Camera


tomaka:
The teapot example was supposed to have an FPS camera. It's not totally implemented because I was lazy, but that was the project initially.

Sorry if that wastes your work, but what about modifying `support/mod.rs` instead?

Roxxik:
I don't think it'd be wasted work, because i can just rewrite it to fit in `support/camera.rs`

PR #1392, opened 2016-01-04
Fixing camera.rs of the example

Roxxik:
I'm not sure if this is the right thing, but it seems legit to me
especially changing the direction, because being at position z=1 and looking down the positive z axis shouldn't let you see the teapot that is located at the origin

### DIFF
--- a/examples/support/camera.rs
+++ b/examples/support/camera.rs
@@ -18,7 +18,7 @@ impl CameraState {
         CameraState {
             aspect_ratio: 1024.0 / 768.0,
             position: (0.1, 0.1, 1.0),
-            direction: (0.0, 0.0, 1.0),
+            direction: (0.0, 0.0, -1.0),
             moving_up: false,
             moving_left: false,
             moving_down: false,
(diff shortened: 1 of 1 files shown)


tomaka:
I used these pages as a reference:

https://www.opengl.org/sdk/docs/man2/xhtml/gluLookAt.xml
https://www.opengl.org/sdk/docs/man2/xhtml/gluPerspective.xml

So I don't really understand what's going on.

Roxxik:
first of I fixed the default value of the camera, becuase if you look from 0, 0, 1 in the direction 0, 0, 1 then you'll never see the origin because your are looking away from it. and after that i removed some negations:
- 1. The perspective matrix negates z (the other changes to the perspective doesn't change anything)
- 2. The view matrix negates forward (default that is z)


and because there is a negation in the view matrix, the result looks wrong, while the coordinates are right
just look at the output of `println!("{:?}", camera.get_position);` (put this somewhere in the main loop)
and with the default values you see that pressing forward moves you in the positive z direction, but it should be the negative z direction


so to sum it all up, i just swapped z and -z. 
It's right I alread heard that too. The OpenGL docs have the default world space with positive z in front and the device space with negative z. Because this is quite confusing (and we have to calculate our own matrices either way) we can do it the most straight forward way.

tomaka:
Ok, let's merge this.

I think the calculation is still incorrect, but since it's even more incorrect in the master branch, it can't harm to merge this.

Issue #1391, opened 2016-01-03
How to implement picking/selection using glium


Roxxik:
I'm currently working on a picking feature. Skimming through the glium code I found nothing hinting at direct support for it. So i want to implement my own and am not very sure how to do it,
I generally read about 3 options to picking:
- 1. Color picking: draw each object/primitive in a different color (no light/fog/particles) and just read the pixel value under your cursor
- 2. Depth only rendering: for each object/primitive: render object, read depth-buffer at mouse position, clear depth buffer. pick minimum depth object
- 3. ray casting. this one has not much to do with glium

for 1 I think i could render a scene this way, but I don't know how to render to a not on screen buffer and read a pixel value
for 2 basically the same with just depth: how to render to an only depth buffer and how to read values from it.

If someone would hint me at the functions to look at to do this I'd be happy to implement is as an example.

tomaka:
I created a gist by copy-pasting from my project: https://gist.github.com/tomaka/d2708687a7947f02d630

The first file is a raw copy-paste, I didn't touch anything.
As you can see, at line 125 the `draw` function takes the picking texture and builds a framebuffer with it at lines 141 or 147. The shader writes to that output at line 88. The value that is written comes from the vertex attributes.

The second file shows how to use it. I have a the picking texture and a pixel buffer of size 1. At each frame, the one pixel of the buffer is read to determine what was hoving the mouse at the previous frame. Then we adjust the texture to the dimensions of the window if necessary. It then calls `ModelsSystem::draw` by passing the picking texture. Finally it copies the pixel that is under the mouse to the buffer so that it will be read at the next frame.

tomaka:
For the record, if you want to draw to just a depth buffer, you can use `SimpleFrameBuffer::depth_only`. This function was just added by #1387 and is not yet on crates.io.

tomaka:
And before you ask, you can't draw at the same time to the window and to the picking texture. You will have to draw the colors to a texture as well, then copy that color texture to the window with the `blit` function for example.

Roxxik:
I'll got a [basic version](https://github.com/Roxxik/glium/blob/picking/examples/picking.rs) up and working, it's not cleand-up yet, especially the picking_target builds way too many intermediate objects and the picking assumes that id = index + 1, that is not verified anywhere. But before cleaning up I'd like to know, why the picking is not working on the edges of the model and sometimes even if the mouse is not on the model.

Roxxik:
[...]

PR #1393, opened 2016-01-04
Picking example

Roxxik:
How to implement picking in glium #1391


Roxxik:
There might be a off by one for the pixel values, not yet sure about this, needs to be tested.

This leads to objects getting picked while the mouse is one pixel right of them or objects not getting picked while the mouse is on the leftmost pixel of the object.

Roxxik:
There was an issue if two objects touched on the screen and you'd move the mose from one object to the other, both were green.

This is now fixed in d62f21fb22182d0822b61d333e174213baf937c0

tomaka:
Let's merge this, thanks!

PR #1398, opened 2016-01-08
Add additional GL context information to capabilities struct

bitshifter:
Also added info example which outputs context information based on GLFW's glfwinfo test. Example output looks like:

```
OpenGL context verson: 4.3.0 - Build 10.18.15.4256
OpenGL context flags: debug
OpenGL profile mask: compatibility
OpenGL robustness strategy: none
OpenGL context renderer: Intel(R) HD Graphics 4600
OpenGL context vendor: Intel
```

### DIFF
--- a/examples/info.rs
+++ b/examples/info.rs
@@ -1,0 +1,55 @@
+#[macro_use]
+extern crate glium;
+
+fn main() {
+    use glium::{Api, CapabilitiesSource, DisplayBuild, Profile, Version};
+    use glium::backend::Facade;
+    use glium::glutin;
+
+    // building the display, ie. the main object
+    let display = glutin::WindowBuilder::new()
+        .with_visibility(false)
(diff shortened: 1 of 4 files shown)


tomaka:
Perfect, thanks!

tomaka:
Ah I didn't notice you didn't add getters.
The `CapabitiliesSource` trait is a hack around the problems with visibility in Rust. It's not supposed to be used in user code.

bitshifter:
Ah right, I did notice getters on Context. I can add getters there. One question, I added the gl version string, I think it's useful since it has additional information over the Version struct. Maybe get_opengl_version_string() in addition to the existing get_opengl_version()? If yes then I'd probably also use get_opengl_renderer_string() and get_open_gl_vendor_string() for consistency.

tomaka:
Sounds good.

PR #1403, opened 2016-01-13
Subroutines

nicokoch:
*This is a WIP, do not merge yet!*

I needed subroutines for a personal project, so I started implementing them in glium.

If this feature is not desired, it's perfectly fine if you reject this.

So far, I have implemented reflection for subroutine data. This includes all subroutine uniforms and their corresponding candidate subroutines. This is turn means that the "data part" is done, which enables error checking etc. to avoid opengl errors.

The next step is the hard part where I need some pointers though, because I'm not so familiar with gliums internals: Actually drawing.
1. The user can potentially specify the subroutines in a format similar to this: `(ShaderType, "uniform_name") => "subroutine_name"`. How would these subroutine values be handled in the API? As an extra parameter to draw? Or maybe as draw_parameters? (I figured handling subroutine_uniforms the same as uniforms would be dumb, because they are definitely not the same thing)
2. Where do the current values get cached (context?) to avoid unnecessary opengl calls? (This could also be implemented lateron)
3. These caches need to be invalidated when i.e. after calling `glUseProgram()`, how is this handled in glium?

Of course I'm happy about feedback regarding my current implementation too.
Oh, I also added an example program which prints the queried SubroutineData. It gives a nice overview of the data structures.

Thanks in advance

### DIFF
--- a/src/program/reflection.rs
+++ b/src/program/reflection.rs
@@ -1031,7 +1031,7 @@ pub struct SubroutineData {
     pub subroutine_uniforms: HashMap<(String, ShaderStage), SubroutineUniform>,
 }
 
-/// Information about a Subroutine Uniform (except name and shaderstage)
+/// Information about a Subroutine Uniform (except name)
 #[derive(Debug, Clone)]
 pub struct SubroutineUniform {
 
@@ -1057,6 +1057,8 @@ pub struct Subroutine {
     pub name: String,
 }
 
+/// The different stages of the program pipeline.
+#[allow(missing_docs)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
 pub enum ShaderStage {
     Vertex,
@@ -1070,6 +1072,7 @@ pub enum ShaderStage {
 }
 
 impl ShaderStage {
+    /// Converts the `ShaderStage` to its GLenum equivalent
     pub fn to_gl_enum(&self) -> gl::types::GLenum {
         match *self {
             ShaderStage::Vertex => gl::VERTEX_SHADER,
@@ -1095,17 +1098,22 @@ fn get_supported_shader_stages(ctxt: &mut CommandContext) -> Vec<ShaderStage> {
 }
 
 /// Returns the data associated with a programs subroutines.
-/// Returns `None` if subroutines are not supported by the backend.
 pub unsafe fn reflect_subroutine_data(ctxt: &mut CommandContext, program: Handle)
-                               -> Option<SubroutineData>
+                               -> SubroutineData
 {
     if !ctxt.extensions.gl_arb_shader_subroutine {
-        return None
+        return SubroutineData {
+            location_counts: HashMap::with_capacity(0),
+            subroutine_uniforms: HashMap::with_capacity(0),
+        }
     }
 
     let program = match program {
         // subroutines not supported.
-        Handle::Handle(_) => return None,
+        Handle::Handle(_) => return SubroutineData {
+            location_counts: HashMap::with_capacity(0),
+            subroutine_uniforms: HashMap::with_capacity(0),
+        },
         Handle::Id(id) => id
     };
 
@@ -1166,8 +1174,8 @@ pub unsafe fn reflect_subroutine_data(ctxt: &mut CommandContext, program: Handle
             subroutine_uniforms.insert((uniform_name, *stage), subroutine_uniform);
         }
     }
-    Some(SubroutineData {
+    SubroutineData {
         location_counts: location_counts,
         subroutine_uniforms: subroutine_uniforms
-    })
+    }
 }
--- a/src/program/reflection.rs
+++ b/src/program/reflection.rs
@@ -1017,3 +1017,157 @@ fn glenum_to_transform_feedback_mode(value: gl::types::GLenum) -> TransformFeedb
         v => panic!("Unknown value returned by OpenGL varying mode: {}", v)
     }
 }
+
+/// Contains all subroutine data of a program.
+#[derive(Debug, Clone)]
+pub struct SubroutineData {
+    /// Number of subroutine uniform locations per shader stage.
+    /// This is *not* equal to the number of subroutine uniforms per stage,
+    /// because users can use `#layout(location=...)`.
+    pub location_counts: HashMap<ShaderStage, usize>,
+
+    /// The list of all subroutine uniforms of the program stored in a structured way to enable fast lookups.
+    /// A subroutine uniform is uniquely defined by a name and a shader stage.
+    pub subroutine_uniforms: HashMap<(String, ShaderStage), SubroutineUniform>,
+}
+
+/// Information about a Subroutine Uniform (except name and shaderstage)
+#[derive(Debug, Clone)]
+pub struct SubroutineUniform {
+
+    /// The index of the subroutine uniform.
+    /// Needed to query information from the OpenGL backend.
+    pub index: u32,
+
+    /// The location of the uniform.
+    /// This is used to bind subroutines to this subroutine uniform.
+    pub location: i32,
+
+    /// A list of subroutines that can potentially be used with this uniform.
+    pub compatible_subroutines: Vec<Subroutine>,
+}
+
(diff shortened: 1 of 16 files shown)


tomaka:
Nice!

> 2. Where do the current values get cached (context?) to avoid unnecessary opengl calls? (This could also be implemented lateron)
3. These caches need to be invalidated when i.e. after calling glUseProgram(), how is this handled in glium?

As far as I know the current subroutines is per program. So the state would simply be stored in the `Program` struct.

> 1. The user can potentially specify the subroutines in a format similar to this: (ShaderType, "uniform_name") => "subroutine_name". How would these subroutine values be handled in the API? As an extra parameter to draw? Or maybe as draw_parameters? (I figured handling subroutine_uniforms the same as uniforms would be dumb, because they are definitely not the same thing)

That's the tricky point. I don't have an answer to that for now.

tomaka:
There's another feature that will need a serious change in the API: program pipelines. This feature allows you to create "incomplete programs" (for example a program with just a vertex shader and another program with just a fragment shader) and use them when drawing. But each program has its own uniforms.

A good change in the API needs to be found for both subroutines and program pipelines.

tomaka:
I quickly read the specs and I think subroutines can use the same interface as uniforms.

After all the thing whose value is set by the programmer is called a `subroutine uniforms`. Other features like uniform buffers or SSBOs also use the same `uniform!` struct in glium, even though they aren't strictly uniforms either.

nicokoch:
Thanks for the detailed answers. I will try to get a prototype running.

nicokoch:
This last commit contains the subroutine implementation.

Subroutines can be used just like regular uniforms with the `uniform!` macro. To set a subroutine, the user has to give the macro a subroutine name and a corresponding `ShaderStage` as a parameter:

```rust
let uniforms = uniform! {
        Color: ("ColorRed", ShaderStage::Fragment)
};
```

A full example can be seen in `examples/subroutines`.

TODOs:
- Tests
- Documentation

nicokoch:
Added tests.

src/program/reflection.rs
```diff
@@ -1017,3 +1017,164 @@ fn glenum_to_transform_feedback_mode(value: gl::types::GLenum) -> TransformFeedb
...
+        for i in 0..subroutine_count {
+            // Get the name of the uniform
+            let mut uniform_name_tmp: Vec<u8> = vec![0; 64];
+            let mut name_len: gl::types::GLsizei = mem::uninitialized();
+            ctxt.gl.GetActiveSubroutineUniformName(program, stage.to_gl_enum(), i as gl::types::GLuint, uniform_name_tmp.len() as gl::types::GLint,
+                                                   &mut name_len, uniform_name_tmp.as_mut_ptr() as *mut gl::types::GLchar);
+            uniform_name_tmp.set_len(name_len as usize);
+            // Get the location of the uniform
+            // TODO Not sure if this is completely correct. Do we have to append '\0' ?
+            let location = ctxt.gl.GetSubroutineUniformLocation(program, stage.to_gl_enum(), uniform_name_tmp.as_ptr() as *const gl::types::GLchar);
```
tomaka: I think this is wrong. `name_len` contains the length of the string, excluding the null terminator. So when you use `set_len()` above you're removing the null terminator from the vec.

A simple fix is to call `set_len` after calling `GetSubroutineUniformLocation`.

src/uniforms/value.rs
```diff
@@ -962,3 +964,17 @@ impl AsUniformValue for (u64, u64, u64, u64) {
...
 impl_uniform_block_basic!((u64, u64, u64, u64), UniformType::UnsignedInt64Vec4);
+
+impl<'a> AsUniformValue for (ShaderStage, &'a str) {
+    #[inline]
+    fn as_uniform_value(&self) -> UniformValue {
+        UniformValue::Subroutine(self.0, self.1)
+    }
+}
+
+impl<'a> AsUniformValue for (&'a str, ShaderStage) {
```
tomaka: I find it weird that there are two different conventions.
Shouldn't we stick to one?

src/program/reflection.rs
```diff
@@ -1017,3 +1017,164 @@ fn glenum_to_transform_feedback_mode(value: gl::types::GLenum) -> TransformFeedb
...
+    }
+}
+
+fn get_supported_shader_stages(ctxt: &mut CommandContext) -> Vec<ShaderStage> {
+    let mut stages = vec![ShaderStage::Vertex, ShaderStage::Fragment];
+    if ctxt.extensions.gl_arb_tessellation_shader {
+        stages.push(ShaderStage::TessellationControl);
+        stages.push(ShaderStage::TessellationEvaluation);
+    }
+    if ctxt.extensions.gl_arb_geometry_shader4 {
```
tomaka: There are `is_geometry_shader_supported` and `is_tessellation_shader_supported` functions in `program/mod.rs`. Please use these instead.

src/program/reflection.rs
```diff
@@ -1017,3 +1017,164 @@ fn glenum_to_transform_feedback_mode(value: gl::types::GLenum) -> TransformFeedb
...
+        stages.push(ShaderStage::Geometry);
+    }
+    stages
+}
+
+/// Returns the data associated with a programs subroutines.
+pub unsafe fn reflect_subroutine_data(ctxt: &mut CommandContext, program: Handle)
+                               -> SubroutineData
+{
+    if !ctxt.extensions.gl_arb_shader_subroutine {
```
tomaka: Subroutines are in core 4.0. Please also add a check `|| ctxt.version >= &Version(Api::Gl, 4, 0)`.

src/program/reflection.rs
nicokoch: Turns out my current implementation is buggy anyways. We have to check if the program actually has the shaders, not if they are supported. Is `RawProgram::has_tessellation_shaders` safe to use? Can I also add a `has_geometry_shader`?

tomaka: `has_tessellation_shaders` was used in the past but no longer is. The problem is that when you load a program from a binary, you have no way to know which shader stages are used.

nicokoch: So is there any way to know if a program has tesselation and/or geometry shaders?

src/uniforms/value.rs
nicokoch: I'll stick to the `(&str, ShaderStage)` variant, which I think is more intuitive.

src/program/reflection.rs
tomaka: There's a third awkward solution, which is that glium could add a small header when returning a program binary and parse it when reloading.

OpenGL implementations are free to refuse to load any binary for whatever reason they want, so technically it would be a valid thing to do.

nicokoch: I guess that's the best idea.

I can implement this in the same run. My suggestion would be to add a single byte at the beginning of the binary like this:
```
byte & 0b00000001 => has_tess_evaluation
byte & 0b00000010 => has_tess_control
byte & 0b00000100 => has_geometry
```
Failing if any of the first 5 bits aren't 0 (can later be used for additional flags)
Thoughts?

nicokoch:
I just addressed the nitpicks in this latest commit.

Regarding the binary file custom header, I wrote this earlier but it's now gone because I pushed:

I guess that's the best idea.

I can implement this in the same run. My suggestion would be to add a single byte at the beginning of the binary like this:
```
byte & 0b00000001 => has_tess_evaluation
byte & 0b00000010 => has_tess_control
byte & 0b00000100 => has_geometry
```
Failing if any of the first 5 bits aren't 0 (can later be used for additional flags)
Thoughts?

tomaka:
Yeah that's a good format.

nicokoch:
That's done too. This PR is getting kind of big ^.^

tomaka:
Nice!

The new code looks good.
Is the PR ready to be merged?

nicokoch:
Go ahead and merge it. Documentation is not done yet but I can submit that in another PR during the next week.

BTW I reenabled the `program_binary` tests because I think I fixed the bug that caused Mesa to fail.

PR #1407, opened 2016-01-18
Expose safe interface to OpenGL timer queries

nicokoch:
This module provides utilities to profile GPU execution times in a new module `glium::timer`.

This example shows how to use the new functionalities (from the docs):

```rust
// drawing a frame
use glium::timer;
let mut target = display.draw();
let handle = timer::measure(&display, || {
    target.clear_color(0.0, 0.0, 0.0, 0.0);
    target.draw(&vertex_buffer, &index_buffer, &program, &uniforms, &Default::default()).unwrap();
    target.finish().unwrap();
}).unwrap();
println!("GPU needed {:?} to draw", handle.get_measurement().interval());
```

I had this code lying around from ancient times and ported it to glium.
It makes use of `GL_arb_timer_query`


src/timer.rs
```diff
@@ -0,0 +1,225 @@
...
+    ctxt.gl.GenQueries(1, &mut timer_query_id);
+    ctxt.gl.QueryCounter(timer_query_id, gl::TIMESTAMP);
+    timer_query_id
+}
+
+/// Executes the closure and measures the time needed by the GPU
+///
+/// If the GPU is idle during execution, the clock keeps ticking, so it doesn't make a lot
+/// of sense to use this function for CPU-heavy tasks.
+pub fn measure<F, M>(facade: &F, measure: M) -> Result<AsyncTimerHandle, TimerCreationError>
```
nicokoch: I'm not really good with the closure traits. Is `FnOnce` the correct choice?

src/timer.rs
```diff
@@ -0,0 +1,225 @@
...
+                     where F: Facade, M: FnOnce() -> ()
+{
```
nicokoch: I'm not really good with the closure traits. Is `FnOnce` the correct choice?

src/timer.rs
```diff
@@ -0,0 +1,245 @@
...
+    Duration::new(seconds, nanos)
+}
+
+/// Executes the closure and measures the time needed by the GPU.
+/// Any pending GPU operations will not be included.
+///
+/// If the GPU is idle during execution, the clock keeps ticking, so it doesn't make a lot
+/// of sense to use this function for CPU-heavy tasks.
+pub fn measure<F, M>(facade: &F, measure: M) -> Result<AsyncTimerHandle, TimerCreationError>
+                     where F: Facade, M: FnOnce() -> ()
```
nicokoch: I'm not really good with the closure traits. Is `FnOnce` the correct choice?

nicokoch:
I see something similar has already been implemented in [draw_parameters](https://github.com/tomaka/glium/blob/e7cdb2197b983b0d1d52d7b165e4bc0a99dc6193/src/draw_parameters/query.rs), so maybe this won't be very useful.

tomaka:
Isn't that what's already implemented in debug.rs? https://github.com/tomaka/glium/blob/master/src/debug.rs#L119

nicokoch:
Oh you're right.

What do you think about the general api?
I know other query objects are currently handled in draw parameters, but my design would allow profiling things like compute shaders (or other non-draw related GL functionality) too.

tomaka:
Your design works for timestamp queries, but not for other kind of queries because the user can only activate one query of each time simultaneously.

PR #1411, opened 2016-01-22
use num not $num in macros

dten:
otherwise num is unused and results in a warning. assuming i can't remove it all together though for type problems?


tomaka:
Thanks

PR #1426, opened 2016-02-14
Add and document RawImage2d::from_raw_rgb methods

ruud-v-a:
On a side note, the `_reversed()` methods could take a slice or even an `IntoIterator` as they make a copy anyway, but that would be a breaking change.


tomaka:
Thanks!

PR #1437, opened 2016-03-18
use GLObject Trait instead of BufferExt::get_buffer_id

Kerosene2000:
I am sorry if there is errors of implementations, I am not as good as you in Rust, I work hard for it :smiley_cat:

Happy to contribute to this beautiful lib :heart:


src/vertex/buffer.rs
```diff
@@ -54,6 +58,15 @@ impl Error for CreationError {
     }
 }
 
+impl<T: Index> GlObject for VertexBuffer<T> {
```
tomaka: That shouldn't be `T: Index` but just `T`.

tomaka:
Looks good other that the minor problem.

PR #1438, opened 2016-03-18
forgotten uniforms::Buffer GLObject Trait implementation

Kerosene2000:


src/uniforms/buffer.rs
```diff
@@ -19,6 +22,15 @@ pub struct TypelessUniformBuffer {
     buffer: BufferAny,
 }
 
+impl<T: Copy> GlObject for UniformBuffer<T> {
```
tomaka: `<T: ?Sized + Content>`

The bounds should simply be the same as what the main object requires.

tomaka:
I had forgotten about that PR! Thanks!

PR #1447, opened 2016-04-03
Fix a few warnings

ruud-v-a:
There are still a few warnings left that I did not fix:

 * There are some unused results. Calling `unwrap()` would be a behavioural change and it might have a small performance impact.
 * Making types public to avoid the private type in public interface warning causes more warnings than it solves due to the missing documentation warning. It will have to be fixed at some point though, because it will become a hard error.


tomaka:
Thanks!

PR #1452, opened 2016-04-10
Add temporary workaround for CircleCI problem

Bobo1239:


tomaka:
I don't really understand the fix, but thanks.

Bobo1239:
[Explanation & Source](https://github.com/rust-lang/cargo/issues/2429#issuecomment-191370340) | [Cargo issue](https://github.com/rust-lang/cargo/issues/2078)

It's only needed in the first command because after that, everything is cached (each command (`-`) runs in a separate shell so the key isn't available anymore in the second command)

PR #1454, opened 2016-04-17
fixed warnings that were going to be made into errors next release

vadixidav:
If for some reason panic on failure to copy a rect was not desired, I can modify the PR with a more appropriate solution (such as ignoring an error). This is primarily to eliminate the annoying warnings that will turn into errors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -382,7 +382,7 @@ trait UniformsExt {
 ///
 /// Blocks and subroutines are not included.
 #[derive(Copy, Clone, Debug)]
-enum RawUniformValue {
+pub enum RawUniformValue {
     SignedInt(gl::types::GLint),
     UnsignedInt(gl::types::GLuint),
     Float(gl::types::GLfloat),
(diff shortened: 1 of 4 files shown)


vadixidav:
It seems the examples don't use the new MouseMoved that expands the tuple, so the test failed. It isn't due to the PR though.

tomaka:
> If for some reason panic on failure to copy a rect was not desired, I can modify the PR with a more appropriate solution (such as ignoring an error). This is primarily to eliminate the annoying warnings that will turn into errors.

Actually it's wrong to panic, and it's wrong to ignore the errors too. The right solution is propagate them, but this requires breaking the API.

src/lib.rs
```diff
@@ -382,7 +382,7 @@ trait UniformsExt {
 ///
 /// Blocks and subroutines are not included.
 #[derive(Copy, Clone, Debug)]
-enum RawUniformValue {
+pub enum RawUniformValue {
```
tomaka: Could you add `#[doc(hidden)]` so that it doesn't show up in the docs?

bnewbold:
@vadixidav if you rebase on `master`, tests should pass

vadixidav:
Just noticed there was no merge. Feel free to merge #1458. The only relevant change here is the pub extern crate issue if the other things don't need to be fixed, and since that PR is already up its probably easier that way.

PR #1455, opened 2016-04-18
examples/picking: trivial fix for MouseMoved API

bnewbold:
Travis and CircleCI tests have been failing for glium because (as far as I can tell) of a small change in the glutin MouseMoved API.
This trivial commit fixes the problem.


tomaka:
Thanks

PR #1470, opened 2016-05-12
Support texture upload to texture arrays.

mheistermann:
Support texture upload to 2d texture arrays.
Fix off-by-one when x==width.

### DIFF
--- a/src/texture/any.rs
+++ b/src/texture/any.rs
@@ -836,6 +836,24 @@ impl<'a> TextureAnyMipmap<'a> {
         self.depth
     }
 
+    /// Computes a tuple (width, height, depth) of mipmap dimensions,
+    /// using 1 for unused dimensions.
+    ///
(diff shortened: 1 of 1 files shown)


tomaka:
Hmm, I would find it surprising that `y` corresponds to array layers for 1D arrays.

mheistermann:
Can't find anything about this in the glTexSubImage2D documentation, but at least https://www.opengl.org/sdk/docs/man/html/glCopyTexSubImage2D.xhtml states

> When target is GL_TEXTURE_1D_ARRAY then the y coordinate and height are treated as the start slice and number of slices to modify.

So i think it would make sense to just behave the same way.

mheistermann:
If you prefer, I could add a seperate layer range parameter, just let me know.

tomaka:
No I think that's fine.
You just need to rebase your branch over master and I'll merge.

mheistermann:
Rebased

PR #1471, opened 2016-05-12
book/tuto-08: mention GLSL version bump to v150

### LINKED ISSUE #1311: Tutorial 08 doesn't run
While running step 8 of the tutorial, I ran into the following error:

    thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: CompilationError("0(12) : error C7532: global function inverse requires \"#version 150\" or later\n0(12) : error C0000: ... or #extension GL_ARB_gpu_shader5 : enable\n")', ../src/libcore\result.rs:731
    An unknown error occurred

kyrylo:
Fixes #1311 (Tutorial 08 doesn't run)

### DIFF
--- a/book/tuto-08-gouraud.md
+++ b/book/tuto-08-gouraud.md
@@ -22,7 +22,7 @@ to the surface, then the surface should be dark.
 We are going to do this calculation one per fragment, in our fragment shader. The brightness of
 each pixel will need to be equal to `sin(angle(surface, light))`. If the light is perpendicular,
 the angle is `pi/2` radians and the brightness is `1`. If the light is parallel, the angle is `0`
-and the brightness is `0`. 
+and the brightness is `0`.
 
 The question is: how do we know the angle between the surface and the light? This is where
 normals come into play.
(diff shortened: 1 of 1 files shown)


book/tuto-08-gouraud.md
```diff
@@ -42,10 +42,12 @@ If the light is *parallel* to the normal, then the surface should be bright. And
 ## In practice
 
 The main part of the calculation will be done in the fragment shader. However we need to modify
-the vertex shader first, in order to pass the normal's data to the fragment shader.
+the vertex shader first, in order to pass the normal's data to the fragment shader. In addition
+to this, also note that some hardware doesn't support some functions that we're going to use in
```
tomaka: It's not "some hardware that doesn't support some functions". It's just that GLSL v140 itself doesn't support these functions.

kyrylo: How about:

> In addition to this, we need to specify a newer version of GLSL, since v140 doesn't support the functions we're going to use. To make the vertex shader work, we need at least GLSL v150.

tomaka: That would be good!

kyrylo:
PTAL

tomaka:
Thanks

PR #1477, opened 2016-05-18
Unignore some tests that work now

nicokoch:
The subroutine tests were previously causing segfaults due to bugs in mesa, but they got fixed after I reported them some time ago.

The bug in `persistent_mapping_forget_then_read` I could no longer reproduce, and travis is not complaining either.


tomaka:
Thanks!

PR #1485, opened 2016-06-09
Fix typos in tuto 01/02

passy:
Two more small fixes. :)

---

This change is [Reviewable](https://reviewable.io/reviews/tomaka/glium/1485)


IvanUkhov:
It’d be great if you could also fix line 86 which contains “a attribute.” Thanks!

PR #1487, opened 2016-06-12
Trivial kcov gitignore

bnewbold:
Has kconv (coverage) skip /tests/ code (and also any code living in /target/, which currently includes glutin files for the glium build), and adds some trivial entries to gitignore.


tomaka:
Aren't the patterns local? For example why would you need to exclude /usr/bin?

Also I'm against adding editor stuff in .gitignore.

bnewbold:
Thanks for reviewing! I've pushed some updates.

I've removed the gitignore changes.

Having re-run `kcov` locally, the '/usr/lib/' exlude pattern does seem
unnecessary; i'm not sure what I added that before. Removed. I've added a
comment explicitly justifying the remaining exclude patterns to the commit.

PR #1488, opened 2016-06-12
Check for GLES 3 in a few places. This is not comprehensive.

### LINKED ISSUE #1482: OpenGL ES 3.0+ requires integer signed/unsigned 2d textures...
... But glium only checks for OpenGL 3.0+, and thus thinks they're unsupported.

DiamondLovesYou:
Fixes #1482.


src/program/raw.rs
```diff
@@ -335,7 +335,8 @@ impl RawProgram {
         let value = unsafe {
             match self.id {
                 Handle::Id(id) => {
-                    assert!(ctxt.version >= &Version(Api::Gl, 2, 0));
+                    assert!(ctxt.version >= &Version(Api::Gl, 2, 0) ||
+                            ctxt.version >= &Version(Api::GlEs, 2, 0));
```
tomaka: It looks like only GLES 3.0 supports GetFragDataLocation. GLES 2.0 should return None.

PR #1498, opened 2016-07-10
Fixed erroneous TransformFeedbackNotSupported error

Benjamin-L:
When creating a shader with transform feedback varyings, glium will always give the error `TransformFeedbackNotSupported`. This is because of a typo in `src/program/program.rs`. The test for transform feedback still passed because it is designed to silently succeed if the computer doesn't support transform feedback.


tomaka:
Oh, thanks!

Issue #1310, opened 2015-10-21
gpgpu example doesn't run


am0d:
I get the following output when I try executing the compute shader example:

    thread '<main>' panicked at 'called `Result::unwrap()` on an `Err` value: CompilationError("0(5) : error C0000: syntax error, unexpected \';\', expecting \"::\" at token \";\"\n")', ../src/libcore\result.rs:731
    Process didn't exit successfully: `target\debug\examples\gpgpu.exe` (exit code: 101)

I can't tell if this is due to my machine setup, or a bug in the compute shader program, since I've never worked with compute shaders before.

luxvio:
Same here, never used it ... but `glxinfo` shows this:

```text
OpenGL vendor string: NVIDIA Corporation
OpenGL renderer string: GeForce GTX 960/PCIe/SSE2
OpenGL core profile version string: 4.4.0 NVIDIA 355.11
OpenGL core profile shading language version string: 4.40 NVIDIA via Cg compiler
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile
OpenGL core profile extensions:
OpenGL version string: 4.5.0 NVIDIA 355.11
OpenGL shading language version string: 4.50 NVIDIA
OpenGL context flags: (none)
OpenGL profile mask: (none)
OpenGL extensions:
OpenGL ES profile version string: OpenGL ES 3.1 NVIDIA 355.11
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.10
OpenGL ES profile extensions:
[snip]
    GL_ARB_compressed_texture_pixel_storage, GL_ARB_compute_shader, 
    GL_ARB_compute_variable_group_size, GL_ARB_conditional_render_inverted, 
    GL_NV_blend_square, GL_NV_command_list, GL_NV_compute_program5, 
    GL_ARB_compressed_texture_pixel_storage, GL_ARB_compute_shader, 
    GL_ARB_compute_variable_group_size, GL_ARB_conditional_render_inverted, 
    GL_NV_blend_square, GL_NV_command_list, GL_NV_compute_program5, 
```

tomaka:
@luxvio That's a weird problem.

Could you add this code here: https://github.com/tomaka/glium/blob/master/src/program/shader.rs#L211

```rust
println!("{:?}", ctxt.get_version());
println!("{:?}", ctxt.get_extensions());
```

...and check what is printed?

luxvio:
Ok, that passage ended up like this:

```rust
[...]

PR #1500, opened 2016-07-14
Adjust compute shader example

abadd:
This addresses issue #1310 where the compute shader example,
gpgpus.rs, does not run on certain setups.  With the most
recent up to date nVidia driver the example would have no issues
but on an older driver, capable of running OpenGL 4.5 and over a
year old, the example would panic due to a compilation error of
the shader.


tomaka:
Thanks

PR #1504, opened 2016-07-25
Replace clock_ticks with std::time in examples/support/mod.rs

serprex:


serprex:
circleci seems an unrelated failure

PR #1510, opened 2016-08-10
Fix typo

baskerville:


tomaka:
Hmm, sorry but I don't see any change.

ozkriff:
'a attribute' -> 'an attribute'

tomaka:
Ah thanks!

Issue #1502, opened 2016-07-21
can't read float data from a texture


darkstalker:
Seems that the only method to read from a `Texture2d` is `fn read<T>(&self) -> T where T: Texture2dDataSink<(u8, u8, u8, u8)>` [1] , and that's hardcoded to u8, so I can't read a texture with float components (`UncompressedFloatFormat::F32F32F32F32`) without a significant precision loss. 
Is that possible to implement? afaik it should've the equivalent of calling `glGetTexImage(texture, 0, GL_RGBA, GL_FLOAT, &pixels)`. The idea is to read the unclamped raw float values written by a shader.

[1] http://tomaka.github.io/glium/glium/texture/texture2d/struct.Texture2d.html#method.read

Vtec234:
I think it would be possible to fix by rewriting `read<T>` to be more generic - could be `read<T, P>(&self) -> T where T: Texture2dDataSink<P>`. The current value of `Texture2dDataSink<(u8, u8, u8, u8)>` forces a `ClientFormat` of `ClientFormat::U8U8U8U8` which matches this line in the `ops` module: https://github.com/tomaka/glium/blob/9fd866b5cf6de2a2415786b24844713a29359c2f/src/ops/read.rs#L271 and results in _always_ calling `glReadPixels(..., GL_RGBA, GL_UNSIGNED_BYTE)`, which as you rightly remarked causes all kinds of problems.
I'll try to work on some workaround, but given that the `glium` is imo extremely complicated, I don't promise anything.

Boscop:
@darkstalker Did you try `unchecked_read_to_pixel_buffer` or `unchecked_read`?
Btw, which format would you recommend for fetching pixel data with 10-bit color-depth?

PR #1513, opened 2016-08-18
Make read and read_to_pixel_buffer generic over pixel formats and fix nightly build

Vtec234:
Currently, there is no way to sanely read from a texture whose internal format is not unsigned integral data, as explained in #1502. The fix for this appears to be very simple, as `PixelValue` is already implemented for a bunch of tuples such as `(f32, f32, f32, f32)`, and the internal functions used in `read` also contain valid dispatches for all pixel formats into the enum supplied to `glReadPixels`. I haven't tested this thoroughly, but it does work for `F32F32` and I don't see a reason why it wouldn't work for other formats.

Another thing this PR does is to fix a build failure with the most recent nightly Rust due to `RawUniformValue` being private, but present in a public interface.

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -316,6 +316,7 @@ fn build_texture<W: Write>(mut dest: &mut W, ty: TextureType, dimensions: Textur
             use texture::{{TextureCreationError, Texture1dDataSource, Texture2dDataSource}};
             use texture::{{Texture3dDataSource, Texture2dDataSink, MipmapsOption, CompressedMipmapsOption}};
             use texture::{{RawImage1d, RawImage2d, RawImage3d, CubeLayer}};
+            use texture::pixel::PixelValue;
 
             use image_format::{{ClientFormatAny, TextureFormatRequest}};
             use image_format::{{UncompressedFloatFormat, UncompressedIntFormat}};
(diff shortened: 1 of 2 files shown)


tomaka:
The reason why this isn't implemented is that `glReadPixels` is not guaranteed to support anything else but U8U8U8U8. The proper fix would be to check the OpenGL version and stuff, but that's not implemented yet.

Vtec234:
I see. However, given that this functionality is extremely useful, maybe it could be added with some commentary on unstability/lack of support guarantees?

Alternatively, we could limit the accepted `PixelValue`s to ones specified as far back af OpenGL 1.5. The table of formats there (Table 3.6 in 1.5 spec) doesn't allow `R, G`, so types like `(f32, f32)` would have to be excluded, but others (1/3/4 components, any primitive) should work.

tomaka:
> as it looks U8U8U8U8 shouldn't be the only supported format.

The problem is OpenGL ES 2.0, which is only guaranteed to support two formats: U8U8U8U8 and an implementation-specific format which you can query.

What you could do, though, is add `unchecked_read` and `unchecked_read_to_pixel_buffer`, which do the same as the current functions but accept any format and are unsafe. Then change `read` and `read_to_pixel_buffer` to be simple wrappers around their unchecked equivalent

Vtec234:
Alrighty, created `unchecked` variants. I may give checking context compatibility a shot, as it should be a matter of simply calling `self.0.get_context().get_opengl_version()` and checking that against the specs' `format` tables in something like `is_format_supported_in_version(v: &Version, f: ClientFormat)`. As long as you are fine with doing this on every `read`, that is. As for unsupported formats, it could either panic (too harsh?) or return `Err`, but that would be a breaking change of the interface from `T` to `Result<T, GliumReadError>`.

tomaka:
> checking that against the specs' format tables in something like is_format_supported_in_version(v: &Version, f: ClientFormat).

That's not how it works. A format can be supported for writing but not reading.
The formats that you can read are entirely different and are not implemented at all.

Vtec234:
Thanks for merging!

As for the formats, in that case this: https://github.com/tomaka/glium/blob/master/src/ops/read.rs#L201 would have to be flagged as a bug/TODO, since it uses a `ClientFormat`, which is client-side for texture uploading, to obtain parameters for a read operation.

PR #1517, opened 2016-08-20
Make use of a faster hasher in many places

### LINKED ISSUE #946: Replace hash functions with faster ones
When the hash API is stabilized in Rust.

jonas-schievink:
Applying the first 4 commits noticably improved performance when we developed [Plantex](https://github.com/OsnaCS/plantex/), the last one didn't make too much of a difference.

Closes #946

### DIFF
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -39,7 +39,7 @@ impl VertexAttributesSystem {
     #[inline]
     pub fn new() -> VertexAttributesSystem {
         VertexAttributesSystem {
-            vaos: RefCell::new(HashMap::new()),
+            vaos: RefCell::new(HashMap::with_hasher(Default::default())),
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/vertex_array_object.rs
```diff
@@ -17,12 +16,13 @@ use gl;
...
 use version::Api;
 use version::Version;
+use utils::fnvmap::*;
 
 /// Stores and handles vertex attributes.
 pub struct VertexAttributesSystem {
     // we maintain a list of VAOs for each vertexbuffer-indexbuffer-program association
     // the key is a (buffers-list-with-offset, program) ; the buffers list must be sorted
-    vaos: RefCell<HashMap<(Vec<(gl::types::GLuint, usize)>, Handle), VertexArrayObject>>,
+    vaos: RefCell<FnvHashMap<(Vec<(gl::types::GLuint, usize)>, Handle), VertexArrayObject>>,
```
tomaka: Please leave the regular `HashMap` here, as it is likely faster than the fnv hashmap.

tomaka:
On the principle the change is good. But when it comes to coding style I think I prefer having explicitly `HashMap<..., BuilderHasherDefault<FnvHasher>>` and `HashMap::with_hasher(Default::default())` everywhere instead of creating a module for that.

jonas-schievink:
ping @tomaka

PR #1522, opened 2016-09-11
Rename some of the examples and ignore Intellij files

berardino:
1 - Remove hyphens from examples files to make Intellij happy 
https://github.com/intellij-rust/intellij-rust/issues/664

2 - Ignore Intellij files in gitignore


tomaka:
As a general statement, I don't want to adapt my code to a specific IDE.
If the IDE doesn't support hyphens in file names, then it's probably the IDE that should be fixed and not the file names.

berardino:
well yes, I agree and in fact I have also opened an issue on intellij-rust plugin. But this would have been nice to have considering that this changes don't really tight your code to any specific IDE.

PR #1526, opened 2016-09-19
Fix image orientation in screenshot example

baskerville:
The triangle should point upward given its coordinates.

But glReadPixels returns pixels starting from the bottom left corner, so we need to flip vertically in order to get a correct image.


tomaka:
Tanks!

tomaka:
Thanks*

I got a server error from github when trying to merge :o

Issue #1430, opened 2016-03-02
`Surface::clear` doesn't clear sub-rects correctly on `Frame`.


dylanede:
If you look at https://github.com/tomaka/glium/blob/master/src/lib.rs#L1131 you'll see that the `rect`argument of `clear` isn't even used, and a `None` is passed to the corresponding parameter in `ops::clear`. Is this intended, or an oversight? It appears that other implementations of `clear` do pass through the `rect` argument through correctly, instead of ignoring it.

tomaka:
Oh, right that's a bug.

PR #1528, opened 2016-09-22
Pass optional Rect on Frame.clear()

vrav:
Simple edit for #1430 observed in March. The Rect was there, but the call was passing None.


tomaka:
Thanks

PR #1550, opened 2016-12-02
Make examples run on Mac OS X

SirVer:
Also remove some unused symbols.

With these changes all examples but the ones listed below run on my Mac OS X 10.11.5 Early 2013 MacBook Pro. I cannot say if they also output the correct visuals since I did not read all code.

The following do not work (properly?):
- gpgpu: Crashes with 'not yet implemented', in glutin-0.6.2/src/api/cocoa/headless.rs:60
- picking: runs, but only displays a black window.
- screenshot-asynchronous: runs, but does not create a screenshot. Crashes on
  window close with: 'The `Frame` object must be explicitly destroyed by
  calling `.finish()`', src/lib.rs:1221

### DIFF
--- a/examples/deferred.rs
+++ b/examples/deferred.rs
@@ -6,14 +6,12 @@ extern crate image;
 use glium::glutin;
 use glium::index::PrimitiveType;
 use glium::Surface;
-use glium::DisplayBuild;
 use std::io::Cursor;
 
 mod support;
 
 fn main() {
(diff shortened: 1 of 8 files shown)


SirVer:
The shader version update is required for the `inverse` function that is undefined otherwise.

tomaka:
Thanks!

PR #1558, opened 2017-01-08
Publish 0.16.0

bvssvni:
Updated glutin and image version. See https://github.com/PistonDevelopers/conrod/pull/902#issuecomment-271123242


tomaka:
Thanks

PR #1562, opened 2017-01-21
Add support for vertex normalization.

### LINKED ISSUE #1313: Normalize
I can't find how to normalise how to normalize the vertex attributes (the `normalized` option of [glVertexAttribPointer](https://www.opengl.org/wiki/GLAPI/glVertexAttribPointer)). Is it missing ?

Nitori-:
Normalization enables better packing of vertex data. This closes #1313.

This is probably not the cleanest way to do it, but it's the best I could think of.

My first approach was to have a wrapper struct that marked that the data should be used with normalization. The problem is rust doesn't guarantee the memory layout of 'newtype' style structs.

I also would have liked if this could have used the same "arm" as the old implement_vertex. But I couldn't figure out how to make that work.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -122,12 +122,50 @@ macro_rules! implement_vertex {
                                 let dummy: &$struct_name = unsafe { ::std::mem::transmute(0usize) };
                                 attr_type_of_val(&dummy.$field_name)
                             },
+                            false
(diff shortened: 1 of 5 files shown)


tomaka:
Nice approach. Thanks.

PR #1563, opened 2017-01-24
Fix UniformValue::is_usable_with for depth textures

lspitzner:
There does not appear to be a shadow `UniformType` for `DepthTexture3d`.


tomaka:
Thanks

Issue #1566, opened 2017-01-29
Creating an f32 texture fails where an u8 one works


pedrocr:
I have a Vec<f32> with width * height * 3 elements (interleaved RGB). If I create a texture by first converting it to u8 it works:

```rust
let dims = (imgbuf.width as u32, imgbuf.height as u32);
let mut img = vec![0 as u8; imgbuf.data.len()];
for (o, i) in img.chunks_mut(1).zip(imgbuf.data.chunks(1)) {
  o[0] = (i[0]*255.0).min(255.0).max(0.0) as u8;
}
let raw_image = glium::texture::RawImage2d::from_raw_rgb_reversed(img, dims);
let img = glium::texture::Texture2d::new(&display, raw_image).unwrap();
```

but if I just create the texture from a copy of the original array:

```rust
let dims = (imgbuf.width as u32, imgbuf.height as u32);
let img = (*imgbuf).data.clone();
let raw_image = glium::texture::RawImage2d::from_raw_rgb_reversed(img, dims);
let img = glium::texture::Texture2d::new(&display, raw_image).unwrap();
rawid = Some(image_map.insert(img));
```

I get a 'Texture data size mismatch' panic in src/texture/any.rs:147 from the Texture2d::new call.

Am I using the API wrong somehow? It seems the only thing that changes is if I'm passing in Vec < u8 > or Vec < f32 >.

tomaka:
Glium supposes that each element in the Vec in a single pixel.
If you want to change this behavior you have the possibility to manually implement the `Texture2dDataSource` trait.

pedrocr:
from_raw_rgb(a)_*() already takes care of the interleaved stuff in the u8 case, why not in the f32 case?

pedrocr:
Fixed by #1568

PR #1568, opened 2017-01-30
Make RawImage2d::from_raw_rgb* accept more than u8

pedrocr:
Possible fix for #1566. The from_raw_rgb*() functions do advertise accepting Vec<T> but then set the texture format to U8U8U8 and U8U8U8U8 effectively making them only work with Vec<u8> data. Instead set the correct format for u8/i8/u16/i16/u32/i32/f32 by constraining T to a trait that specifies the format and implementing the trait for those types.


tomaka:
Thanks. I wonder why I put only U8U8U8 here in the past.

PR #1569, opened 2017-01-31
Take reference instead of Vec in RawImage2d::from_raw_rgb(a)_reversed() to avoid extra copy

pedrocr:
The from_raw_rgb(a)_reversed() functions do a copy but take a Vec<T> and thus may require copying if the calling code needs the original data back. Change them to a &[T] which works fine and avoids the issue.


pedrocr:
Any objection to this? Seems like a straightforward enough optimization

tomaka:
Thanks

PR #1570, opened 2017-01-31
Added categories to Cargo.toml

mmmbacon:
Added two categories to cargo.toml. First PR, please be gentle :)


tomaka:
Unfortunately the category names have to match these values here: https://crates.io/category_slugs

Also glium is not really "GUI". Instead it should be "rendering::graphics-api" I think.

tomaka:
Thanks

Issue #1572, opened 2017-02-07
VertexBuffer's invalidate method is either confusingly documented or busted


Sushisource:
Hi there - really enjoying glium, just getting started with it. I'm not sure if this is just a documentation issue or an actual bug. 

The docs here:
https://tomaka.github.io/glium/glium/struct.VertexBuffer.html#method.invalidate

Seem to imply that if I 

    buffer.invalidate()
    buffer.write(some_verts_which_is_shorter_than_buffer_size)

It will be OK with that and I can populate just a portion of the buffer. However in practice the call to write() still panics b/c the data is shorter than the buffer.

Is this behavior expected? Should the  docs change? If so, how should I efficiently approach a scenario where the amount of verts I'm trying to write is changing every frame?

Thanks!

tomaka:
`invalidate()` is just a hint for the OpenGL implementation. It has no visible effect except that it's as if you wrote random data to the buffer. So yes, the behavior is expected.

You should be able to write to a portion of the buffer by doing `buffer.slice(0 .. num_vertices).unwrap().write(...)`.

Sushisource:
OK, makes sense. My real issue here isn't actually writing just a portion of the buffer, but rather needing to grow the buffer as I'm going. What I'm rendering is an animation of drawing a boundary around some pixels, and I'm drawing it as I'm computing it, using a LineStrip.

So, my thought was I could allocate a large enough buffer, and render only the part I've populated with meaningful data.

I'm not an opengl expert by any means but IIRC you would accomplish that in raw opengl by passing a number of elements to render to the draw call. What's the equivalent mechanism for that in glium?

Thanks!

PR #1578, opened 2017-03-08
Update docstring for invalidate()

Sushisource:
References issue #1572 

Simple change to the docstring that would've helped me avoid my confusion in that issue. Hopefully this is a help!


tomaka:
Thanks, but I don't think the documentation of `invalidate()` should mention anything about rendering or give an example about how to write to the buffer.

I think it should just say something like "This does not change the size of the buffer. See `write` for an example to write to a subset of the buffer.", and add the write example to `write`.

PR #1580, opened 2017-03-21
[WIP] Update to the new `EventsLoop` glutin/winit API

mitchmindtree:
This change is in anticipation of tomaka/glutin#864 which updates to winit's new `EventsLoop` based API. See the linked PR for details.

Now that winit `Window`s require an `EventsLoop` argument to their
`build` method, the current `DisplayBuild` trait does not work well with
the `glutin` backend. To compensate for this, the `DisplayBuild` trait
has been implemented for `Window` itself rather than `WindowBuilder`,
allowing the user to build their glutin `Window` separately from
building their glium `Display`.

`build`, `build_debug` and `rebuild` fns have been added to remove the
need for uses to import the `DisplayBuild` trait.

One problem that arose while making this change is that, now that
`rebuild_glium` no longer has access to the `WindowBuilder` for the new
window, it can no longer insert the `.with_shared_lists` builder method.
I see a couple possible alternatives for this:

1. Rely on the user calling
`.with_shared_lists(display.get_window().unwrap())` when building the
new window that is to be used for rebuilding the display.
2. Add a `.share_lists(&other_window)` method to `glutin::Window`.
3. Something else?

@tomaka I haven't updated the book or `Cargo.toml` yet as I thought it best to get your thoughts on everything first, but I'm happy to do so once you are happy with the changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ default = ["glutin"]
 unstable = []       # used for benchmarks
 
 [dependencies.glutin]
-version = "0.7.1"
+version = "0.8"
 features = []
 optional = true
 
--- a/examples/blitting.rs
+++ b/examples/blitting.rs
@@ -5,18 +5,15 @@ extern crate glium;
 extern crate image;
 
 use std::io::Cursor;
-use glium::{DisplayBuild, Surface};
-
-use glium::glutin;
+use glium::{glutin, Surface};
 
 mod support;
 
 fn main() {
     // building the display, ie. the main object
-    let display = glutin::WindowBuilder::new()
-        .with_vsync()
-        .build_glium()
-        .unwrap();
+    let events_loop = glutin::EventsLoop::new();
+    let window = glutin::WindowBuilder::new().with_vsync().build(&events_loop).unwrap();
+    let display = glium::build(window).unwrap();
 
     // building a texture with "OpenGL" drawn on it
     let image = image::load(Cursor::new(&include_bytes!("../tests/fixture/opengl.png")[..]),
@@ -53,14 +50,17 @@ fn main() {
         dest_texture.as_surface().fill(&target, glium::uniforms::MagnifySamplerFilter::Linear);
         target.finish().unwrap();
 
+        let mut action = support::Action::Continue;
+
         // polling and handling the events received by the window
-        for event in display.poll_events() {
+        events_loop.poll_events(|event| {
             match event {
-                glutin::Event::Closed => return support::Action::Stop,
+                glutin::Event::WindowEvent { event: glutin::WindowEvent::Closed, .. } =>
+                    action = support::Action::Stop,
                 _ => ()
             }
-        }
+        });
 
-        support::Action::Continue
+        action
     });
 }
--- a/examples/deferred.rs
+++ b/examples/deferred.rs
@@ -13,12 +13,14 @@ mod support;
 fn main() {
     use cgmath::SquareMatrix;
 
-    // building the display, ie. the main object
-    let display = glutin::WindowBuilder::new()
+    let events_loop = glutin::EventsLoop::new();
+    let window = glutin::WindowBuilder::new()
         .with_dimensions(800, 500)
-        .with_title(format!("Glium Deferred Example"))
-        .build_glium()
+        .with_title("Glium Deferred Example")
+        .build(&events_loop)
         .unwrap();
+    // building the display, ie. the main object
+    let display = glium::build(window).unwrap();
 
     let image = image::load(Cursor::new(&include_bytes!("../tests/fixture/opengl.png")[..]), image::PNG).unwrap().to_rgba();
     let image_dimensions = image.dimensions();
@@ -358,14 +360,17 @@ fn main() {
         target.draw(&quad_vertex_buffer, &quad_index_buffer, &composition_program, &uniforms, &Default::default()).unwrap();
         target.finish().unwrap();
 
-        // polling and handling the events received by the window
-        for event in display.poll_events() {
-            match event {
-                glutin::Event::Closed => return support::Action::Stop,
-                _ => ()
-            }
-        }
+        let mut action = support::Action::Continue;
 
-        support::Action::Continue
+        // polling and handling the events received by the window
+        events_loop.poll_events(|event| {
+            match event {
+                glutin::Event::WindowEvent { event: glutin::WindowEvent::Closed, .. } =>
+                    action = support::Action::Stop,
+                _ => (),
+            }
+        });
+
+        action
     });
 }
--- a/examples/displacement_mapping.rs
+++ b/examples/displacement_mapping.rs
@@ -9,12 +9,10 @@ use glium::Surface;
 mod support;
 
 fn main() {
-    use glium::DisplayBuild;
-
     // building the display, ie. the main object
-    let display = glutin::WindowBuilder::new()
-        .build_glium()
-        .unwrap();
+    let events_loop = glutin::EventsLoop::new();
+    let window = glutin::WindowBuilder::new().build(&events_loop).unwrap();
+    let display = glium::build(window).unwrap();
 
     let image = image::load(Cursor::new(&include_bytes!("../tests/fixture/opengl.png")[..]),
                             image::PNG).unwrap().to_rgba();
@@ -218,14 +216,17 @@ fn main() {
                     &Default::default()).unwrap();
         target.finish().unwrap();
 
+        let mut action = support::Action::Continue;
+
         // polling and handling the events received by the window
-        for event in display.poll_events() {
+        events_loop.poll_events(|event| {
             match event {
-                glutin::Event::Closed => return support::Action::Stop,
+                glutin::Event::WindowEvent { event: glutin::WindowEvent::Closed, .. } =>
+                    action = support::Action::Stop,
                 _ => ()
             }
-        }
+        });
 
-        support::Action::Continue
+        action
     });
 }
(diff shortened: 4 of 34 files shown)


geomaster:
Hey, thanks for doing this, I was looking forward to seeing the new API in glium. Unfortunately I didn't really have the time to perform the update myself, upon seeing how much work it'd be, and I also missed this PR.

What's the current status of this PR? I'm just asking since there doesn't seem to have been any activity during ~2 months. Are you waiting for @tomaka's feedback?

If there's any way I can help get this finished and merged, please let me know, I'll be happy to.

mitchmindtree:
Yeah mainly I'd like to get some general feedback and some thoughts on the `rebuild_glium` part of my original comment, otherwise I think all that remains is a rebase and to update the tutorial examples. The review doesn't necessarily have to be from tomaka as I know he's busy with his own stuff atm, just someone with some glium experience would be great. That way if tomaka doesn't have time to review himself, he can at least be more confident in merging it knowing that multiple users have given the all clear.

geomaster:
I'm far from a glium expert, but I've read some of the code and fiddled around, so I can take a look, but heads up that there's a good chance I'll miss something :) I'll get around to reviewing in a few days.

geomaster:
Hey @mitchmindtree, sorry for the delay, I was a bit swamped.

First of all, I took a look at all your changes, and they're looking good! I was a bit concerned about adding `glium::build` and friends in `lib.rs`, as they felt detached, but after thinking about it for a while I have to admit that doing `glium::build(window)` feels very natural. The placement of those functions in the root of the crate makes sense, as they are an entry point to the whole library. But perhaps that's something that @tomaka should have a final say in, as it's a cosmetic API change (small, but still). As for me, I'm all for ergonomics, and so I'm all good with the change.

As for the bit about list sharing, I'll admit that the issue looks tricky. While I'm not zealously opposed to it, I don't like your first solution. Relying on the user to do something that the API should know about and do is error prone, and I think that a solution where the library retains the same behavior should be prioritized. That said, I'm not sure how common of a use-case rebuilding the context is, so you could argue that the people doing that know what they're doing, and will read and understand the documentation (where this should probably be explained, should we go that route).

Your second option sounds better from an API standpoint. However, I have some concerns about its viability. From what I've been able to reckon by digging through glutin's code, the story with list sharing is as follows:

1. On Windows (WGL), glutin [queries for the existence](https://github.com/tomaka/glutin/blob/0dd337ca67773f9cde55c6d6a875f2de93fa8ee9/src/api/wgl/mod.rs#L226) of `wglCreateContextAttribsARB`, which allows the user to share lists with a window as part of the context creation. If `wglCreateContextAttribsARB` is not found, glutin falls back to calling `wglCreateContext` and then `wglShareLists`.
2. On Linux-based platforms, EGL seems to [not have this feature implemented](https://github.com/tomaka/glutin/blob/d915964731657e6936ddbed6cce823e0bd3223b9/src/api/egl/mod.rs#L168) and GLX [looks like it does](https://github.com/tomaka/glutin/blob/d915964731657e6936ddbed6cce823e0bd3223b9/src/api/glx/mod.rs#L369).

Adding a `.share_lists(&window)` would then be possible only on Windows, through calling `wglShareLists`, since neither EGL nor GLX provide a function to share OpenGL objects after the context is created. However, the docs on this function seem to be sparse. The best resource I was able to find is [this](https://www.khronos.org/opengl/wiki/Platform_specifics:_Windows#wglShareLists), which suggests that *there is a chance that the call will fail if any objects are created in either of the contexts before the call*. This sounds strange and would suggest that even with the current implementation, the lists might not actually be shared, as clearly we've created lots of OpenGL objects in the existing context by the time we get to doing a `rebuild`.

Given the almost non-existent and conflicting documentation, and the requirement that no objects are created in either/one of the contexts, I'd say that a method such as `.share_lists(&window)` would be hard, if not impossible, to use correctly,  Also, adding it would prohibit us from enabling this behavior on Linux-based systems, as both GLX and EGL give us a chance to share lists with an existing context only on context creation, but not afterwards. So this solution is also a bit problematic. 

The root cause seems to be that, now that `Window` is `DisplayBuild`, `rebuild` is simply too late in the pipeline to dictate something that should ideally be known at `Window` construction time. Perhaps use the first solution you proposed, but add an alternative method which would take a `WindowBuilder` and an `EventsLoop`, construct the window with lists shared, and return the display? The user can then always get to the actual window by calling `.get_window().unwrap()`. Perhaps call the method `rebuild_glium_and_window`? I can think of nothing better :(

As an aside, I'm a bit confused about why glium documentation says that [*the old context **must** have lists shared with the new one*](https://tomaka.github.io/glium/glium/backend/struct.Context.html#method.rebuild), but EGL seems to not implement this feature, leading me to believe that doing a rebuild on an EGL platform would break the application doing that. All this is kind of pushing my understanding of OpenGL context management to its humble limits, so I can only count on you or @tomaka helping me make sense of all this :)

mitchmindtree:
Thanks a lot for the review @geomaster - just wanted to mention that I'll get around to responding to this soon but I'm just trying to address some upstream changes first ([here](https://github.com/tomaka/winit/pull/191)).

mitchmindtree:
@geomaster I've finally had the chance to sit down and read through your response properly. Thanks a lot for digging into the `rebuild_glium` stuff! I agree the options aren't amazing, however I'm currently proposing [a redesign for glutin](https://github.com/tomaka/glutin/pull/897) which may make it easier to tackle this.

## glutin redesign and `glium::rebuild`

The basic gist of the redesign is to separate the `Window` from the OpenGL `Context` in order to 1. make maintaining glutin easier and 2. give users more granular control over the way they use windows and contexts. In glutin's case, this means the `glutin::Display` type would contain individual fields for both the `glutin::Context` and the `winit::Window`, rather than just a single `glutin::Window` type which wrapped both the winit window and the GL context specific code. This should allow us to rebuild the inner `glutin::Context` inside the `glium::rebuild` function, ensuring that the lists are indeed shared as they were in the original API. The revised function might look like this:

```rust
glium::rebuild(window, context_builder, &display);
```

@geomaster I'm currently after a review for the glutin redesign at tomaka/glutin#897 - if you get a chance to leave your thoughts there that would be greatly appreciated! 

## `DisplayBuild` unnecessary

One thing I'd also like to see is the removal of `DisplayBuild` trait entirely. The trait itself is not required as a bound on any functions or types throughout all of glium, and seems to exist solely as an attempt to unify the way in which displays for different backends are built. Realistically, different backends (e.g. glutin vs glfw vs sdl2) are inherently always going to have different needs in their construction process, and trying to provide one way of building all possible `Display`s only acts as a hinderence for maintainers of those different backends.

E.g. we've already seen this with the introduction of the `EventsLoop` in winit, where it resulted in no longer being able to implement `DisplayBuild` for `glutin::WindowBuilder`, instead moving the impl to `glutin::Window` itself (which caused other problems, e.g. `rebuild_glium as discussed in this PR). This would require being changed *yet again* if the glutin redesign at tomaka/glutin#897 is accepted.

Instead, I think we're better off removing the `DisplayBuild` trait and the crate-level functions I have added in favour of adding public methods directly to the `GlutinFacade` type (which is the `Display` type when the `glutin` feature is active) for `build`, `build_debug`, `build_unchecked` and `rebuild`. These would naturally be feature gated behind the `glutin` feature along with the `Display` type itself. The *really* important trait is the `Backend` trait, which acts as the glue between glium and the opengl context providers. `DisplayBuild` on the other hand seems to be purely for aesthetic purposes and does not seem to be worth the maintenance issues.

sharazam:
bump? What happened to this?

sharazam:
It's broken for the latest winit version. Somehow `EventsLoop::interrupt(&self)` got removed from winit. @tomaka @mitchmindtree Is this intentional? The docs say that you should call `interrupt()`, but there is no `interrupt()` anymore in winit.

This is basically the problem. `winit` updates so frequently, you have to basically build it against the `master` branch. Right now, glium / glutin work if you use `winit v.0.6.4`, but it doesn't work against `winit:master`.

```
error[E0599]: no method named `interrupt` found for type `winit::EventsLoop` in the current scope
  --> /home/felix/Development/glutin/src/platform/linux/wayland.rs:79:32
   |
79 |         self.winit_events_loop.interrupt()
```

mitchmindtree:
> Somehow EventsLoop::interrupt(&self) got removed from winit

@sharazam See tomaka/winit#191

mitchmindtree:
ATM I'm waiting on more feedback on tomaka/glutin#897 before updating this PR.

sharazam:
Just wanted to know. I originally didn't want to do anything to the event loop. I only wanted to investigate why `.with_decorations(false)` doesn't work when using glium, but works when using winit standalone. Somewhere in the middle, the "false" is going lost (on X11).

But as it is right now, I can't go on with this because I don't know if this may be fixed in `v.0.6.4`. Right now I am disabling the interrupt() method on my repo because it's unclear if / how it would stay. Just so that I can go on with this.

mitchmindtree:
I you get a chance to review or comment on tomaka/glutin#897 that might help to get this PR moving faster 😊 the main thing stalling progress is the lack of people reviewing, making it hard for tomaka to know if he can merge or not

geomaster:
@mitchmindtree I'll have the time in the following couple of days to take a look at that PR. Thanks!

mitchmindtree:
I made [a branch](https://github.com/glium/glium/compare/master...mitchmindtree:glutin_redesign?expand=1) from this PR that updates for the glutin redesign (see tomaka/glutin#897).  I'll merge that branch into this PR once/if the glutin redesign lands.

mitchmindtree:
Closing this in favour of #1601, which is basically an updated version of this that adapts to the glutin redesign at tomaka/glutin#901.

PR #1582, opened 2017-03-24
Added support for OpenGL primitive index restarting - fixes #180

### LINKED ISSUE #180: Implement primitive restart index
Low priority.

sharazam:
This PR adds the missing implementation for enabling GL_PRIMITIVE_RESTART_FIXED_INDEX. For compatibility with GL ES 3.0, only fixed restart index will be supported. The unit test works by drawing two lines and seeing if they are connected (in a successful test, they should not be connected, but rather be two seperate lines). Adds a public field to the draw parameters.

If there are stylistic issues / formatting errors or questions / code review, please notify me. I also fixed some missing / unused imports, otherwise the tests would not compile.

### DIFF
--- a/src/draw_parameters/mod.rs
+++ b/src/draw_parameters/mod.rs
@@ -68,6 +68,7 @@ use version::Version;
 use version::Api;
 
 use index::PrimitiveType;
+use index::IndexType;
 
 use QueryExt;
 use CapabilitiesSource;
@@ -378,6 +379,12 @@ pub struct DrawParameters<'a> {
     /// Since this is purely an optimization, this parameter is ignored if the backend doesn't
     /// support it.
     pub primitive_bounding_box: (Range<f32>, Range<f32>, Range<f32>, Range<f32>),
+    
+    /// If enabled, will split the index buffer (if any is used in the draw call) 
+    /// at the MAX value of the IndexType (u8::MAX, u16::MAX or u32::MAX) start a new primitive
+    /// of the same type ("primitive restarting"). Supported on > OpenGL 3.1 or OpenGL ES 3.0 - 
+    /// otherwise, enabling this has no effect.
+    pub primitive_restart_index: bool,
 }
 
 /// Condition whether to render or not.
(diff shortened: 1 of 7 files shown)


src/draw_parameters/mod.rs
```diff
@@ -378,6 +379,12 @@ pub struct DrawParameters<'a> {
     /// Since this is purely an optimization, this parameter is ignored if the backend doesn't
     /// support it.
     pub primitive_bounding_box: (Range<f32>, Range<f32>, Range<f32>, Range<f32>),
+    
+    /// If enabled, will split the index buffer (if any is used in the draw call) 
+    /// at the MAX value of the IndexType (u8::MAX, u16::MAX or u32::MAX) start a new primitive
+    /// of the same type ("primitive restarting"). Supported on > OpenGL 3.1 or OpenGL ES 3.0 - 
+    /// otherwise, enabling this has no effect.
```
tomaka: If this is not supported, a `FixedIndexRestartingNotSupported` error is returned (which is good). So the last sentence here is wrong.

src/draw_parameters/mod.rs
```diff
@@ -882,3 +891,25 @@ fn sync_primitive_bounding_box(ctxt: &mut context::CommandContext,
         ctxt.state.primitive_bounding_box = value;
     }
 }
+
+fn sync_primitive_restart_index(ctxt: &mut context::CommandContext,
+                                enabled: bool)
+                                -> Result<(), DrawError>
+{
+    if ctxt.version >= &Version(Api::Gl, 3, 1)   || ctxt.version >= &Version(Api::GlEs, 3, 0) ||
```
tomaka: Please add a `TODO` comment saying to use `GL_PRIMITIVE_RESTART` if possible if `GL_PRIMITIVE_RESTART_FIXED_INDEX` is not supported.

tomaka:
Thanks, looks good after the two minor fixes.

sharazam:
Done - should be able to merge. As a note - the test `get_transform_feedback_varyings` failed with the message: 

`0:5(2): error: varying variables must be of base type float in GLSL 1.10")', /glium/tests/shaders.rs:401`

I didn't touch that part, however. Not sure if this could be an issue.

tomaka:
Thanks

PR #1584, opened 2017-03-31
clamp light attenuation to avoid negative outputs

blevin:
while working my way through the enjoyable glium examples i noticed that the lights in the deferred test were not adding together as i'd expect.  turns out the attenuation was going negative, causing the yellow light to subtract from the red and green ones.  a minor thing but easy to fix...

before:
before

with fix:
with_fix

thanks for glium!


tomaka:
Thanks

PR #1593, opened 2017-05-15
Provide current project state in README

Thomspoon:
Since the project is no-longer maintained, it'd be beneficial to have that be the first thing people see when deciding to use Glium.


tomaka:
Thanks!

PR #1594, opened 2017-05-24
Fix SAMPLES query for gles

TimNN:
The [OpenGL ES 3.0 Specification](https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf) states that:

> An `INVALID_ENUM` error is generated if *internalformat* is not color-, depth- or
stencil-renderable; if *target* is not `RENDERBUFFER`; or if *pname* is not `SAMPLES` or
`NUM_SAMPLE_COUNTS`.

(Note that this is different from what the OpenGL or ARB Extension Specification say).


tomaka:
Thanks

PR #1595, opened 2017-05-26
Clarify the current state of glium's development

mitchmindtree:
The current writing reads as though *noone* currently maintains glium which is not the case. This aims to clarify the situation a little.


mitchmindtree:
Does this README change suit you @tomaka?

PR #1596, opened 2017-05-26
Fix invalid texture formats for OpenGL ES 3.0

TimNN:
According to [The OpenGL ES 3.0 Spec](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glTexImage2D.xhtml#id-1.6.11.1) (and my own experiences) some of the unsized internal formats are not available on OpenGL ES (which is in contrast to the situation for [OpenGL 4.0](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glTexImage2D.xhtml#id-1.6.13.1)).

Btw, I'm not sure if my fix is the "nicest" way of writing that -- if you have different preferences let me know.

### DIFF
--- a/src/image_format.rs
+++ b/src/image_format.rs
@@ -1654,6 +1654,8 @@ pub fn format_request_to_glenum(context: &Context, format: TextureFormatRequest,
         None => false,
     };
 
+    let is_es = version.0 == Api::GlEs;
+
     Ok(match format {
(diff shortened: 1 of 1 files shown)


est31:
Hi @TimNN and thanks for the PR. Sorry for giving it a look only now. Could you please rebase so that I can merge? Thank you!

PR #1598, opened 2017-06-15
Fix links in README broken by org move

Tom9729:
Updated the Travis and github.io links in the readme; the old links were broken with the tomaka -> glium move.

CircleCI and Coveralls are still broken.


tomaka:
Thanks!

PR #1603, opened 2017-07-18
Fixedindexrestarting

tl8roy:
Change the logic of `sync_primitive_restart_index()` so that if the OpenGL version is less than `3.0` and `primitive_restart_index` is false, an error message won't be generated


tl8roy:
Note that this fixes the issue I had here: https://github.com/PistonDevelopers/conrod/issues/1022

tomaka:
Thanks

Issue #532, opened 2015-02-19
Consider creating a "glium" organization


tomaka:
I don't want to give the impression this is _my_ project and only _mine_, so a "glium" github organization is maybe better.

On the other hand, I don't see it as absolutely necessary.

palodequeso:
Well, I will say on the extraordinarily rare occurrence that you're not around to fix something that broke with the nightly, it would be nice to have a small group of people that can also fix things.

TyOverby:
Would this organization also own Glutin?

tomaka:
@palodequeso Right. My personal rule is to grant the collaborator status to all people who have contributed to the core of the library. Once people are collaborators, I encourage them to also review and merge PRs.

I've received some fixes in the documentation, updates for rust, and your deferred shading example, but no-one has modified anything in the actual behavior of glium yet.
I'm thinking of being a bit more loose for glium, because I don't see as too good to have a "big" project with only one person.

@TyOverby That's a good question. I don't want to give the impression that glutin becomes tied to glium, they should remain totally independant. On the other hand it's hard to justify creating an organization for glium and not for glutin. But I don't want to create an organization for glutin.

casey:
I just read [the glium post mortem](https://users.rust-lang.org/t/glium-post-mortem/7063/10), and I think that this might be a good idea.

If this repo was transferred to a organization account, people would be less likely to annoy you with questions.

Also, the move would be a good opportunity to make clear that although contributors might review and merge PRs, or occasionally implement new features, the project is abandoned, practically speaking, and is looking for maintainers.

jhasse:
What should be the organization name? https://github.com/glium seems to be taken. https://github.com/rust-glium ?

mitchmindtree:
Just thought I'd mention that I would be interested in being a part of an organisation along these lines if you are still considering the idea. Although my overall OpenGL knowledge is very limited, I will probably continue to contribute bug fixes and basic API updates (e.g. reflecting changes in winit) as I continue to use glium in my own projects. IMO glium is still the most thorough, accessible and idiomatic set of rust bindings to OpenGL that exist and I'll probably continue to use it until that is no longer the case (or if vulkan becomes accessible on macos, or if I happen to stop needing GUI in rust which I doubt will happen any time soon).

bvssvni:
[...]

PR #1604, opened 2017-07-19
Remove deprecation notice from README.md

jhasse:
Since #532 has been fixed, this can be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,17 +4,6 @@
 
 [![](http://meritbadge.herokuapp.com/glium)](https://crates.io/crates/glium)
 
-## Note to current and future Glium users: 
-
-Glium is [no longer actively developed by its original
-author](https://users.rust-lang.org/t/glium-post-mortem/7063). That said, PRs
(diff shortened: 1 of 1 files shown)


rennis250:
Any reason why this has been skipped?

rennis250:
?

tomaka:
Glium is still not really maintained, so I don't see why this text should be removed.

jhasse:
You said that "My personal rule is to grant the collaborator status to all people who have contributed to the core of the library. And once people are collaborators, I encourage them to also review and merge PRs."

I think once some people start reviewing and merging PRs someone from them could take over maintaining.

porky11:
But I think, this part of the message can be removed, since it is in an own organisation
```
If you use Glium and would be interested in seeing it moved to its own organisation, you can express your interest here.
```

paulirotta:
Yes, @tomaka does outstanding work. If his focus remains elsewhere, it might ease his burden if there is someone else who can help with glium PRs. Vulkan is still sadly not supported on Intel devices in a meaningful way so...

PR #1614, opened 2017-08-16
Update dependencies and fix warnings

lilianmoraru:
I saw that some dependencies were not updated for years, so it is probably the time :)

I left 2 warnings(unhandled `Result` - makes me very uncomfortable thinking that the error case was not handled but all those panicky functions should probably return `Result`) because those were not panicking before, so that would of been a breaking API change.


tomaka:
Thanks!

PR #1617, opened 2017-08-25
Fixed segfault in release mode for implement_vertex! macro

fschutt:
Well, the title, basically. My application was crashing in release mode, but not in debug mode. So I wondered why. Debugging lead me to the section where the offsets of the fields get calculated. The reason seems to be that the `mem::transmute` operations seem to cause undefined behaviour because the compiler optimizes certain things out. This leads to a segfault on the (old code) Line 113, because the `mem::transmute` acts on a null pointer. 

This issue only happened in release mode, never in debug mode. If an actual value gets created (with `mem::zeroed`), the compiler doesn't optimize things out and you can actually calculate the correct offset. The code basically subtracts the pointers in order to get the offset of the field.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -108,10 +108,15 @@ macro_rules! implement_vertex {
                         (
                             Cow::Borrowed(stringify!($field_name)),
                             {
-                                let dummy: &$struct_name = unsafe { ::std::mem::transmute(0usize) };
-                                let dummy_field = &dummy.$field_name;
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks! Although it's a bit hacky to put an object on the stack just for that, I don't really see any alternative solution.

lilianmoraru:
@tomaka Any plans to publish a new version on `crates.io`?
For example, this fix seems important, the documentation link is broken now on `crates.io`(it is fixed in `Cargo.toml`), etc...

PR #1628, opened 2017-10-17
Update README to include Organization Path

JustinRyanH:
I noticed the README didn't have the correct circle url to the glium organization. I figured I'd fix this.


JustinRyanH:
After looking into the file struct I realized that we don't really have a config.yml for circle. Should we disable the `glium/glium` project and delete the badge?

PR #1631, opened 2017-10-18
Redirect book urls to the glium/glium repo

JustinRyanH:
Had a previous version of this up, but there was a panic from previous debugging session


tomaka:
Thanks for the PR, but you put `glutin/glutin` and `glutin/glium` at several places, while it's `tomaka/glutin` and `glium/glium`.

JustinRyanH:
@tomaka Good catch. I'll definitely be more thorough next time I push a pull request. I just pushed a change that should take care of the mistakes I made.

Issue #1629, opened 2017-10-17
Question: Any Reason why AsUniformValue  wasn't implemented as From<T>?


JustinRyanH:
I was digging through the Uniform code and I noticed that we use` AsUniformValue` for type conversions. I'm curious if there is a reason glium went with that instead of implementing a bunch of `From<T>` . Would there be any resistance to adding `From<T>` equivalents to the UniformValue enum? I don't think we'd need to deprecate the existing usage.

mitchmindtree:
I imagine that the implementation is currently this way as this code pre-dates the addition of `From` and `Into` in the std, though I could be wrong!

One difference to note is that the `AsUniformValue` conversion method  takes `&self` rather than `self`, but I suppose we could get around this by implementing `From` for references to the types instead, e.g. `From<&'a T>`? Would be nice to hear @tomaka's thoughts briefly before going ahead with this in case there is some special case we're missing.

tomaka:
> I imagine that the implementation is currently this way as this code pre-dates the addition of From and Into in the std, though I could be wrong!

Yes, I think that's the case if I remember correctly.

I think `From` is strictly superior to `AsUniformValue`.

JustinRyanH:
thanks for the response. I'll get together a pull request

tomaka:
Don't both with implementing `From` for `&'a T`. The fact that `AsUniformValue` takes a `&self` instead of a `self` is also a mistake I think.

JustinRyanH:
@tomaka  @mitchmindtree  I've started a pull request to handle the current issue, however I would love to have feedback. I've made an small change to the AsUniformValue api, but I think by the time this PR is ready to merge into master that can hopefully the fixed. https://github.com/JustinRyanH/glium/pull/2

tomaka:
You should make `as_uniform_value()` takes a `self` instead of a `&self`. Don't bother with borrowing. It will make things much simpler as well.

JustinRyanH:
dropping the `&'a self` and `&self` reference is causing a lot of the pain with `Fence` and `Alloc` for buffer. I think I could refactor those into a state that would allow me to do that, but I have some questions on the intent of their structure.  

I've noticed `Buffer<T>` has Drop, so it makes it rough to with moving it's values into other objects. I'm curious if we should instead implement a Tuple struct of (`Alloc`, `Fence`) that is passed around as a `std::rc::Rc ` between Buffers and that can handle the cleanup of those two when there are no buffers referencing that combination.

[...]

PR #1632, opened 2017-10-27
Replace From/Into Traits to replace AsUniformValue

JustinRyanH:
## Goal
Replace AsUniformValue with `From<T> for UniformValue` equivalent

### Why?
I preferred the intent of Glium over other existing OpenGL implementations. However, I want to use some of the newer Stable Rust features such as Macros 1.1. This is one of many refactors to make that code more simple. 

### Notable Changes
https://github.com/glium/glium/pull/1632/files#diff-e4e6538764117aa97918d216423e804d
I made `struct Fences` to bypass issues with Inserters and lifetimes. 

### Tests
Other than the new tests for the UniformValue I asserted that examples rendered exactly the same on a MacbookPro on the previous commit as with the current commit. I also looked for Github projects that used `glium = "0.18.*"`. I ran their examples and also assert that the output was the same as the crate 

#### Resolves
https://github.com/glium/glium/issues/1629

### DIFF
--- a/src/buffer/fences.rs
+++ b/src/buffer/fences.rs
@@ -11,9 +11,6 @@ use std::rc::Rc;
 use context::CommandContext;
 use sync::{self, LinearSyncFence};
 
-
-pub type RefFences = Rc<Fences>;
-
 /// Contains a list of fences.
 #[derive(Clone)]
 pub struct Fences {
--- a/src/buffer/fences.rs
+++ b/src/buffer/fences.rs
@@ -6,28 +6,31 @@ This module handles the fences of a buffer.
 use smallvec::SmallVec;
 use std::cell::RefCell;
 use std::ops::Range;
+use std::rc::Rc;
 
 use context::CommandContext;
 use sync::{self, LinearSyncFence};
 
+
+pub type RefFences = Rc<Fences>;
+
 /// Contains a list of fences.
+#[derive(Clone)]
 pub struct Fences {
-    fences: RefCell<SmallVec<[(Range<usize>, LinearSyncFence); 16]>>,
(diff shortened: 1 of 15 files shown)


src/sync.rs
```diff
@@ -74,6 +74,7 @@ impl Drop for SyncFence {
 /// The fence must be consumed with either `into_sync_fence`, otherwise
 /// the destructor will panic.
 #[must_use]
+#[derive(Clone)]
```
JustinRyanH: This likely shouldn't have clone anymore.

src/buffer/fences.rs
```diff
@@ -6,28 +6,31 @@ This module handles the fences of a buffer.
 use smallvec::SmallVec;
 use std::cell::RefCell;
 use std::ops::Range;
+use std::rc::Rc;
 
 use context::CommandContext;
 use sync::{self, LinearSyncFence};
 
+
+pub type RefFences = Rc<Fences>;
```
JustinRyanH: This isn't used anymore

JustinRyanH:
@tomaka Let me know if you want me to make any changes. I know this is a large pull request, but I'll address any comments ASAP.

tomaka:
I can't review that. Either this PR sits idle forever or I yolo-merge it.

JustinRyanH:
I'm gonna close then for now, so I don't pollute the open PRs.

slightknack:
Can we revisit this PR? It seems like a really useful addition. I don't know how much has changed since it was written, so it might need to be rebased. I'm running into a lot of lifetime issues due to AsUniformValue taking `&self`; glium is an amazing library, and I'd love for this small hole in the API to be patched. \:)

PR #1642, opened 2017-11-19
Attempt to correct three errors in the gpgpu example

avl:
The buffer size is calculated wrong. This means that the example doesn't actually iterate over NUM_VALUES values, but rather NUM_VALUES-3.

### DIFF
--- a/src/uniforms/mod.rs
+++ b/src/uniforms/mod.rs
@@ -3,7 +3,7 @@ A uniform is a global variable in your program. In order to draw something, you 
 give `glium` the values of all your uniforms. Objects that implement the `Uniform` trait are
 here to do that.
 
-There are two primarily ways to do this. The first one is to create your own structure and put
+There are two primary ways to do this. The first one is to create your own structure and put
 the `#[uniforms]` attribute on it. See the `glium_macros` crate for more infos.
 
 The second way is to use the `uniform!` macro provided by glium:


avl:
This is a trivial fix in the gpgpu example in glium. Unless I've completely misunderstood the example, which is ofcourse completely possible! :-)

avl:
Also, found another problem. The code as it was really only calculated the power for every 4th element in the array in the rust data structure. This is because layout std140 implies a stride of 16 bytes for a simple array of float. See https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL) for the full gory details :-)

avl:
Ok, have done some more investigating. Seems the change from std140 to std430 may bring some other problems. Also, the gpgpu example doesn't work with cargo build --release , for me. I'm investigating. I think all fixes in this pull request are kind of "correct", but they may be unmasking other problems which may be worse.

avl:
Ok, I've done some changes in macros.rs, in the code that calculates offsets of fields within structs.
I don't fully understand why it crashed before, or why my fix solves the problem. It may be that it doesn't actually solve the problem, but just makes it not trigger for me on my machine. So this should probably be inspected by an expert, and/or tested on multiple machines, before any merging is done.

tomaka:
The change to the offset macro is definitely good and needed, but I'm not sure about the GPGPU example itself. It's supposed to work as it is, and your PR modifies too much.

avl:
Thanks for your response! Actually, I'm pretty sure the gpgpu-example has never really worked, not in the form it is now. See https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL) , the part about the definition of std140 vs std430 layout.  Quote: "std430: This layout works like std140, except with a few optimizations in the alignment and strides for arrays and structs of scalars and vector elements. Specifically, they are no longer rounded up to a multiple of 16 bytes"

The example uses an array of scalars (of floats), and the GPU will use 16 byte stride for this when std140 is specified.

Also, the number of elements allocated isn't NUM_VALUES either, since the size calculation didn't include the padding. But that doesn't matter if std430 is chosen, since no padding is needed in that case.

The reason it's not immediately obvious is that you can't really tell from the output whether the operation has been performed for every element (because the data is so random). But if the power is changed to a constant 3 (for example), and the input is changed to 2, you'll see that only every 4th element is computed to 8 (2^3), the rest remain as 2.

One change I didn't do, but I think should be, is changing the structs to be repr(C) . As it is now, it's "only luck" that rust doesn't reorder the fields in the structs, which would stop the example from working.

avl:
Maybe I should have made a pull request per fixed problem, instead of grouping them into one big?

tomaka:
> The example uses an array of scalars (of floats), and the GPU will use 16 byte stride for this when std140 is specified.

I'm pretty confident that only `vec4`s are 16 bytes aligned.

avl:
See also:
https://computergraphics.stackexchange.com/questions/4454/opengl-es-3-uniform-buffer-object-with-float-array

avl:
One more source on std140 memory layout:
https://www.safaribooksonline.com/library/view/opengl-programming-guide/9780132748445/app09lev1sec2.html

tomaka:
Yes, sorry for not going into more details.
The correct fix for this example would be to turn the `float values[256];` into `vec4 values[256 / 4];`, but keep the std140 layout.

avl:
Ah, so that the example still remains workable on holder hardware. That sounds great!

I'll modify this pull request the next time I have time to work on this! :-)

avl:
But surely values[256/4]? Since the NUM_VALUES constant is 4096, shouldn't it be values[4096/4]? Or just values[] (unsized type, supported in GLSL).

PR #1647, opened 2017-11-26
Fix gpgpu example so that the glsl actually affects all elements

avl:
Previously the glsl code only acted on every 4th element.

A comment: This is how I understood from tomaka that it should be. I do think I may have misunderstood you, tomaka! Please give me feedback how you would like it to work :-)

### DIFF
--- a/examples/gpgpu.rs
+++ b/examples/gpgpu.rs
@@ -4,6 +4,7 @@ extern crate rand;
 use glium::glutin;
 
 fn main() {
+
     let context = glutin::HeadlessRendererBuilder::new(1024, 1024).build().unwrap();
     let display = glium::HeadlessRenderer::new(context).unwrap();
 
@@ -20,7 +21,7 @@ fn main() {
             void main() {
                 vec4 val = values[gl_GlobalInvocationID.x];
 
-                values[gl_GlobalInvocationID.x] = pow(val, vec4(power,power,power,power));
+                values[gl_GlobalInvocationID.x] = pow(val, vec4(power));
             }
 
         "#).unwrap();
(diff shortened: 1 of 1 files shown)


examples/gpgpu.rs
```diff
@@ -6,28 +6,34 @@ use glium::glutin;
...
-                float values[256];
+                vec4 values[4096/4];
             };
 
             void main() {
-                float val = values[gl_GlobalInvocationID.x];
-                values[gl_GlobalInvocationID.x] = pow(val, power);
+                vec4 val = values[gl_GlobalInvocationID.x];
+                
+                values[gl_GlobalInvocationID.x] = vec4(
```
tomaka: No need to change this line. `pow(val, power);` should work as well.

avl: I didn't think of that! However, now that I tried it it seems it doesn't actually work. The glsl pow() function is only defined for scalars. At least my nvidia 970 drivers on linux don't accept a vec4 as first parameter.

avl: Or maybe both arguments need to be vec4. Didn't try that. Sorry for speaking too quickly!

avl:
I rewrote the pow call so there is only one call for all 4 elements.

avl:
I also rebased my branch ontop of glium master.

examples/gpgpu.rs
```diff
@@ -14,20 +14,21 @@ fn main() {
...
-                float values[256];
+                vec4 values[4096/4];
             };
 
             void main() {
-                float val = values[gl_GlobalInvocationID.x];
-                values[gl_GlobalInvocationID.x] = pow(val, power);
+                vec4 val = values[gl_GlobalInvocationID.x];
+
+                values[gl_GlobalInvocationID.x] = pow(val, vec4(power,power,power,power));
```
tomaka: Should be `vec4(power)` instead of `vec4(power, power, power, power)`.

PR #1649, opened 2017-12-02
Enable double precision vertex attributes.

Sachyco:
I needed to pass double precision vertex attributes, so I just added 4 additional match arms corresponding to double, dvec2, dvec3, and dvec4.


tomaka:
Thanks

PR #1650, opened 2017-12-05
Add repr(C) for Data struct in gpgpu-example

avl:
This struct should be repr(C), since nothing is guaranteed about the struct layout in rust without this directive. And we *are* actually relying on the struct layout ending up identically to the OpenGL std140-layout (which, with our explicit padding, it will if be guaranteed to if we specify repr(C)).


est31:
Good point, thank you! Wish to rebase?

avl:
I'll take a look at it shortly!

avl:
See if this looks right now!

est31:
Yes, it's been merged now. Thanks for your PR!

PR #1656, opened 2018-01-02
Move draw methods into Facade

ashkitten:
This allows programs to use either a `Headless` or `Display` without wrapping them in an enum and using match statements to draw

I wasn't quite sure of the best place to put the method, but `Facade` looked to be the easiest and was common between anything that has a `Context`.


ashkitten:
bump

PR #1658, opened 2018-01-13
fix test get_transform_feedback_varyings

spearman:
Test failed to create program with "OpenGL does not allow varying of
type int" caused by "varying int color;". Changed to float.

Test also specified the wrong id for the attribute.


tomaka:
Thanks

PR #1659, opened 2018-01-15
Update dependencies & add changelog entries

daboross:
Dependencies updated, and noted changes:

- glutin 0.11->0.12: most notably large change, I'm assuming we're going to want to do this. There don't seem to be any incompatible API changes.
- lazy_static 0.2->1.0: same code, just stable version
- smallvec: 0.4->0.6: only changes features we don't use, this might matter for crates depending on us though it shouldn't be a hard migation if they do.
- cgmath: 0.15->0.16: no changes to features we use
- image: 0.16->0.18: substantial changes, but examples using it still work alright.
- rand 0.3->0.4: substantial changes, but examples using it still work
- gl_generator 0.7->0.8: seems to be mostly a bugfix? I'm not too familiar with this.

All the examples that worked previously on my machine still work, and all tests which passed previously still pass, so these changes seem to work. I don't have a Mac or Windows machine to test with, but I assume CI will catch anything obvious.

I went through and re-added changelog entries up to this release mostly for completeness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## Unreleased
+
+ - Updated glutin to version 0.12.
+ - Updated smallvec from version 0.4 to 0.6.
+ - Updated misc internal dependencies and dev-dependencies (lazy_static, cgmath, rand, image, gl_generator).
+
 ## Version 0.14.0 (2016-04-11)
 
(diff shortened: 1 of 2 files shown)


tomaka:
Oh! Thanks a lot for the changelog update!

daboross:
It's no problem- thank you!

PR #1663, opened 2018-01-16
Fix rendering on dpi-scaled screens

### LINKED ISSUE #1657: Offset rendering on Linux since glutin upgrade
Since 5de27b021f13d4eeed8bab43eefcbee0786cf7dd ("update glutin and gl_generator"), on Ubuntu Artful (16.10) with `nvidia-384`, things render off-centre. Here's the `cargo run --example image`:

image

Here's what it should look like. Same machine, but glium at commit 63577da7987923bf8829b826a6c94f7fdff2be61, before the updates. Note the (r) to the right of the L, and the even margin:

image

It's nothing to do with the window size, it happens at all sizes; the discrepancies in the above screenshots are my laziness.

Not super visually significant in this example, but presumably related to https://github.com/PistonDevelopers/conrod/issues/1120 which is awful.

I am still unable to find a pure glutin example that shows the same behaviour.

daboross:
This accounts for a change in what `get_inner_size()` returns which changed from a dpi-scaled size to an absolute pixel size in https://github.com/tomaka/winit/commit/48902297b7bd02fc20f3ae7012074c6b93674980.

I believe this is the best local fix, as get_inner_size() seems to be (as of winit 0.9) the canonical way to get size-in-pixels of a window.

Fixes #1657.

Thanks to @FauxFaux and @spearman for isolating this issue and finding the specific commit that caused it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
  - Updated smallvec from version 0.4 to 0.6.
  - Updated misc internal dependencies and dev-dependencies (lazy_static, cgmath, rand, image, gl_generator).
  - Replaced the `IntoVerticesSource` trait with `Into<VerticesSource>`.
+ - Fixed [rendering bug](https://github.com/glium/glium/issues/1657) on high-DPI screens.
 
 ## Version 0.19.0 (2017-12-11)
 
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks!

daboross:
it's no problem - thank you for including this!

PR #1664, opened 2018-01-17
Regenerate all documentation from master branch.

daboross:
I've used rustdoc 1.23.0 and gitbook 2.6.7 to regenerate all documentation from the current master branch. I deleted all files in the branch, copied output of rustdoc, and then copied output of gitbook into book/.

~I would have used a 2.x version of gitbook to stay the same as the last generated book, but there was an error stopping me from running it on my computer which was fixed in a 3.x version. It seems fairly similar to the older version, the only difference I've noticed is that there's a search bar above the index now.~

~There's also a difference with the module names: in the old book, they have bolded numbers behind them, and in the one I've just rendered, they are simply plain text. I'm not sure how to fix this.~
Edit: fixed to use gitbook 2.6.7.

This feels like it should be a force push so the gh-pages branch doesn't redundant have history, but I don't know how to do that through a pull request.


daboross:
Fixed gitbook 2.*, pushed updated render with that. This should be good to go.

tomaka:
Thanks for the PR. However at this point I think we should just delete the gh-pages branch altogether.

daboross:
Alright. Do you know if that just has to be done by a org administrator?

My main concern is http://glium.github.io/glium/book/ containing outdated information which doesn't work for the current glium release.

PR #1667, opened 2018-01-18
extend DrawParameters with a bitmask to enable clip planes

### LINKED ISSUE #1662: Extend DrawParameters to control GL_CLIP_PLANEi
OpenGL offers the functionality to clip primitives after the vertex stage against arbitrary planes.
Glium should expose controls for turning the clip planes on and off.

Links:
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/gl_ClipDistance.xhtml
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glClipPlane.xml

mkovacs:
Resolves #1662

### DIFF
--- a/src/draw_parameters/mod.rs
+++ b/src/draw_parameters/mod.rs
@@ -76,6 +76,7 @@ use Rect;
 use ToGlEnum;
 use vertex::TransformFeedbackSession;
 
+use std::mem;
 use std::ops::Range;
 
 pub use self::blend::{Blend, BlendingFunction, LinearBlendingFactor};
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks for the PR. I looked at it and think his is deprecated though!

mkovacs:
Thanks for taking a look!

Can you please clarify what feature you think is deprecated, and in what
version of OpenGL?

What would you like to have changed here?

tomaka:
Sorry, I typed the previous comment during my commute.
It looks like this feature belongs to OpenGL 2, was kept in OpenGL 3 for backward compatibility, then was removed in OpenGL 4.

Glium generally wants to support only modern features.

mkovacs:
I'm looking at the version support chart at the bottom of this page:
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/gl_ClipDistance.xhtml

According to that, gl_ClipDistance is supported by all GLSL versions
starting with 1.3 and going all the way up to 4.5. It's likely supported by
all later versions too which the chart doesn't list due to being out of
date.

What's slightly confusing is that there was an old fixed-function feature
controlled by GL_CLIP_PLANEi, which overlapped in functionality with the
one controlled by GL_CLIP_DISTANCEi, and that was deprecated.

Based on the information available to me, the feature this PR aims to
introduce is alive and well in GL 4+.

mkovacs:
Sure thing! Thanks for merging it!

PR #1670, opened 2018-01-21
Release version 0.20.0

daboross:
This releases 0.20.0 with glutin 0.12, and other changes mentioned in changelog.

I can't think of anything else breaking or otherwise to add to this unless https://github.com/glium/glium/pull/1656 is a go. If it is, I think this PR could just be delayed until that's pulled and a changelog entry is added?

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks for taking care of this!
#1656 is really delicate, so we might skip it for now.

PR #1675, opened 2018-02-13
Fix two compiler warnings

b-r-u:
* Warning 1: the type of this value must be known in this context
  (src/texture/any.rs:289:30)
  Give the raw pointer a type to fix warning that will become a hard error in
  future Rust releases.
  See https://github.com/rust-lang/rust/issues/46906

* Warning 2: unused `std::result::Result` which must be used
  (src/context/mod.rs:528:9)
  Unwrap `Result` of `ops::read` just like in other invocations of this method.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -526,7 +526,8 @@ impl Context {
 
         let mut data = Vec::with_capacity(0);
         ops::read(&mut ctxt, ops::Source::DefaultFramebuffer(gl::FRONT_LEFT), &rect,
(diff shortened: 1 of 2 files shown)


francesca64:
> Warning 2: unused std::result::Result which must be used
(src/context/mod.rs:528:9)

https://github.com/glium/glium/pull/1682#issuecomment-375596152

est31:
These are fixed on master. Thanks for the PR!

PR #1682, opened 2018-03-21
chore: bump glutin, gl_generator version

Brooooooklyn:


Cargo.toml
```diff
@@ -34,18 +34,18 @@ unstable = [] # used for benchmarks
 test_headless = []  # used for testing headless display
 
 [dependencies.glutin]
-version = "0.12"
+version = "0.13"
 features = []
 optional = true
 
 [dependencies]
-backtrace = "0.3.2"
```
daboross: Are these necessary? It doesn't seem like we need to force users of `glutin` to upgrade backtrace/fnv if they don't need to - and if we leave the old version, users are free to use a wider range of compatible versions.

daboross: Are these necessary? It doesn't seem like we need to force users of `glutin` to upgrade backtrace/fnv if they don't need to.

If we leave these as is, users are free to use a wider range of compatible versions.

I previously left this as a single comment, but I think it makes more sense as a review.

Brooooooklyn:
@daboross resolved, thanks~

daboross:
Thanks!

I don't have the knowledge of glium internals necessary to review the second change in `context`, but bumping these two versions sounds good to me.

tomaka:
Thanks for the PR.
Unfortunately, bumping the glutin version is a breaking change and requires an entry in the CHANGELOG.
Also, please don't fix that warning. Its presence is on purpose as a reminder that it should be fixed the correct way.

PR #1685, opened 2018-04-11
Update to glutin 0.14 and gl_generator 0.9. Publish 0.21.0.

### LINKED ISSUE #1673: HiDPI broken on OS X by #1663
Since #1663 `Frame::get_dimensions` returns the size of the window in dpi scaled pixels not raw pixels causing rendering errors.

I think this is a bug with `winit`'s zWindow::inner_size_pixelsz returning the number of points not pixels on OS X but I just want to confirm that the expected behaviour of `Frame::get_dimensions` is to return the size in raw pixels.

Issue example:
screen shot 2018-02-08 at 01 31 53

mitchmindtree:
The CHANGELOG has been updated accordingly.

Closes #1682.
Closes #1674.
Closes #1673.
Closes #1672.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## Version 0.21.0 (2018-04-11)
+
+ - Updated glutin to version 0.14. Fixes handling of HiDPI on macOS.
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks!

PR #1691, opened 2018-06-24
[RFC] Image units: initial support

bobtwinkles:
Adds initial support for image units. Using this support requires either ARB_shader_image_load_store, or the a context conforming to at least the 4.2 core profile. I basically hacked this together to support something that I needed but figured it might be worth upstreaming. The code is not in any way exhaustively tested. Any and all comments are appreciated.


bobtwinkles:
A friend of mine was trying to use this branch and ran into a few traps that I'll note in case someone else wants to use this patch. 
 - All image units are configured to try and sample as `R32I` by default, rather than the format of the texture. Use `set_format` on the image unit to work around this
 - You need to explicitly call `memory_barrier` when doing things that require memory barriers (texture readback, multipass, etc.) There doesn't seem to be an easy way to do this automatically without being ultraconservative and thus sacrificing efficiency unnecessarily.

PR #1698, opened 2018-07-15
Add missing assert branches in program/raw.rs

semtexzv:
Manual shader compilation was crashing on OpenGL ES 2.0 context. I copied assert from earlier place in function.


tomaka:
Thanks

PR #1704, opened 2018-08-11
Bump glutin version to 0.18

### LINKED ISSUE #1703: Certain key presses on OSX are recognised as key releases
On OSX, (and perhaps other OSes, but I haven't tested them), certain modifier key presses, `LControl`, `LShift` etc, as recognised as key releases.

This is really just a glutin 0.17.0 problem, and has been fixed in 0.18.0 (I couldn't find an issue for it however), but it doesn't provide a reason for upgrading that dependency.

To replicate, run the following with glium 0.22.0 and press left control with the window focused:
```rust
extern crate glium;

use glium::glutin::*;

fn main() {
    let mut events_loop = EventsLoop::new();
    let _window = Window::new(&events_loop).unwrap();

    let mut running = true;
    while running {
        events_loop.poll_events(|event| if let Event::WindowEvent {event, ..} = event {
            match event {
                WindowEvent::CloseRequested => running = false,
                WindowEvent::KeyboardInput {input: KeyboardInput {state, virtual_keycode: Some(key), ..}, ..} => {
                    let word = match state {
                        ElementState::Pressed => "pressed",
                        ElementState::Released => "released"
                    };

                    println!("Key '{:?}': {}", key, word);
                },
                _ => {},
            };
        });
    }
}
```

expenses:
Fixes #1703

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ unstable = [] # used for benchmarks
 test_headless = []  # used for testing headless display
 
 [dependencies.glutin]
-version = "0.17"
+version = "0.18"
 features = []
 optional = true
 
--- a/src/backend/glutin/mod.rs
+++ b/src/backend/glutin/mod.rs
@@ -247,6 +247,7 @@ unsafe impl Backend for GlutinBackend {
         match self.borrow().swap_buffers() {
             Ok(()) => Ok(()),
             Err(glutin::ContextError::IoError(e)) => panic!("Error while swapping buffers: {:?}", e),
+            Err(glutin::ContextError::OsError(e)) => panic!("Error while swapping buffers: {:?}", e),
             Err(glutin::ContextError::ContextLost) => Err(SwapBuffersError::ContextLost),
         }
     }


unlimitedbacon:
I think that this example should be updated as well.
https://github.com/glium/glium/blob/0f6c1bd2c631203ddd9ae7787868162cdcede6e5/examples/manual-creation.rs#L42-L48

expenses:
@unlimitedbacon Thanks, I forgot to check tests and examples.

malikolivier:
Any plan to merge this?

r4v3n6101:
Required for fix this: https://github.com/tomaka/winit/issues/607
Also fixed this: https://github.com/Gekkio/imgui-rs/issues/150
So, I think, it should be merged as soon as possible.

daboross:
Doesn't look like anything dramatic was done in https://github.com/tomaka/glutin/compare/v0.17.0...v0.18.0. This looks like a good change to me. :+1:

expenses:
Oh wow, tomaka/winit#607 is actually pretty crippling. Can we get this merged please? @tomaka

daboross:
Worth noting that https://github.com/glium/glium/pull/1711 duplicates this PR, but adds a changelog and immediately bumps the glutin version as well. It may be a better candidate?

Both PRs still look good to me, though, and the code changes are identical except #1711 having different error displays for each error variant where this keeps the same for both.

expenses:
@daboross yep #1711 looks good to me

PR #1711, opened 2018-10-24
Update glutin to 0.18

Lycowolf:
Winit 0.16 has a bug that causes crashes on Windows, but if your project uses glium, this can't be fixed because glutin 0.17 depends on winit 0.16. 

This PR resolves this issue by upgrading glutin and fixing an issue with a new error class that emerged.

Also, I tried to fix a compiler warning about unchecked error return value by panicking (IMHO better than ignoring the error completely).

Tested by compiling Conrod examples (where I discovered this problem) and verifying that it no longer crashes.

PR #1715, opened 2018-12-05
Update glutin dependency to 0.19.

mitchmindtree:
Based on @Lycowolf's #1711 branch, but updates the version to 0.19 rather than 0.18. This includes accounting for the recent consolidation of the `glutin::HeadlessContext` type into the `glutin::Context` type. I've updated the glium CHANGELOG with a link to the glutin 0.19 CHANGELOG entry for further details.

Closes #1711.
Closes #1704.

cc @Lycowolf, @daboross

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Change Log
 
-## Version 0.23.0 (2018-10-22)
+## Version 0.23.0 (2018-12-05)
 
- - Updated glutin to version 0.18.
+ - Updated glutin to version 0.19. See the glutin release notes [here](https://github.com/tomaka/glutin/blob/master/CHANGELOG.md#version-0190-2018-11-09).
 
 ## Version 0.22.0 (2018-07-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Version 0.23.0 (2018-10-22)
+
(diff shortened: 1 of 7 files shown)


mitchmindtree:
@daboross sorry to bug you! Would you (or anyone else!) mind reviewing this and giving it the official "Changes Requested" or "Approved" tick so @tomaka can know whether or not it's safe to merge?

daboross:
This looks good to me! I looked over the changes for glutin 0.19.0 and this handling of headless context looks good.

It seems like we could possibly get rid of glium's `Headless`  struct in favor of just allowing a headless regular display, but that could also not work. In any case leaving it as is works, and at worst it's an extra bit of code.

If you felt like it, updating  more dependencies in this release (genmesh, gl_generator, image, obj, rand) could be good. But that doesn't seem critical.

daboross:
https://github.com/glium/glium/pull/1716 now performs the other dependency upgrades besides `glium` - also looks fully safe as far as I can tell.

My recommendation would now be to merge #1716 then this pull request and release `0.23.0`.

mitchmindtree:
@daboross that sounds good to me!

@tomaka could you please merge #1716 and then this?

daboross:
Travis failed but the failure seems spurious? No log or anything, just "errored". Same with master while the #1716 succeeded which makes me think it's some infrastructure failure on travis which is hopefully resolved by now.

PR #1716, opened 2018-12-09
Update some dependencies and bump version

Eijebong:


daboross:
This looks like a good thing - though the glutin upgrade is redundant with #1715 (and a bit with #1716 and #1711). I think upgrading the other dependencies is a good idea, but #1715 implements better handling of the upgraded glutin, and includes CHANGELOG notes for that.

Maybe this could be reworked to just be updating non-glutin dependencies, and merged alongside #1715?

Eijebong:
Oh, I didn't look if other PRs were opened before writing this... I'll take a look at the mentioned PRs and rework mine accordingly

Eijebong:
There we go, should be mergeable alongside #1715

daboross:
Cool, thanks! I'm not an official maintainer so this'll have to wait for @tomaka but it looks good to me.

If you felt like it this could _probably_ be mentioned in a changelog entry, but then again they're all build or dev dependencies so it shouldn't affect users at all. :+1:

PR #1724, opened 2019-01-31
Remove `backend.is_current` assertion

NathanFlurry:
Originally added in [this](https://github.com/glium/glium/commit/1913af102744d8bfd74902d8a8667d72d6f1dd74) commit, we can't always assert that `backend.is_current` is true after `backend.make_current` is called The issue seems to be an issue specifically on the later versions of macOS – likely specifically in Mojave. For more information, see [this](https://github.com/maps4print/azul/issues/50) issue.

### DIFF
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -644,7 +644,8 @@ impl ContextExt for Context {
             let backend = self.backend.borrow();
             if !backend.is_current() {
                 unsafe { backend.make_current() };
-                debug_assert!(backend.is_current());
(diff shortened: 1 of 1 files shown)


fschutt:
At least from what I can see in azul (since this check was removed in azuls version of glium) the check doesn't seem to do anything. The apps seem to run fine, even when `backend.is_current` evaluates to false. So I think it is just unnecessary and should be removed (although I'm not sure why it doesn't work specifically on Mojave, maybe there is some kind of buffering of driver commands going on).

PR #1735, opened 2019-03-09
Translation of the book into Russian

MrRefactoring:
I work on the translation of the textbook into Russian


Melchizedek6809:
Wow, looks great! Are you still interested and could you update the book so it works with glium:master?

Apart from that adding links in the summary.md files between the Russian and English versions might be useful so that people can easily find the different versions.

MrRefactoring:
Unfortunately, I'm afraid now I don't have much desire and time to support this PR. I'm sorry about that.

est31:
Okay, I'll close the PR then. Translations are a difficult topic, on one hand they are a great addition, on the other hand one can't expect from PR authors to update all translations so it's not easy. I don't know how one would do it, probably via having maintainers specific to each translation, and then those maintainers keeping the translated resources up to date.

The alternative to that is to not have the translators within single projects but instead organized around the language instead, so you have one team for russian speakers, one team for spanish speakers, one for chinese, etc. And then those translators would translate resources into their own languages, in their own repositories.

Melchizedek6809:
Sorry to hear that, but yeah translations are tricky since they need to be kept updated,  maybe keep this one open and add a "Help wanted" label? Since it seems that the entire book has been translated already and from what I can tell mainly the code sections need some minor updating.

Although the glutin 0.30 PR would require some more work to keep things up to date.

Would be a shame to see all this hard work not get merged.

est31:
Yeah why not. It will need either a maintainer though who knows russian, or it will have to live in a different repo outside of glium (where it i maintained instead).

Melchizedek6809:
Alright, then let's leave it open and hope that someone picks up from here

PR #1738, opened 2019-04-14
Warning fixes

est31:
Couple of warning fixes.

r? @mitchmindtree @ZeGentzy

### DIFF
--- a/examples/blitting.rs
+++ b/examples/blitting.rs
@@ -5,6 +5,7 @@ extern crate glium;
 extern crate image;
 
 use std::io::Cursor;
+#[allow(unused_imports)]
 use glium::{glutin, Surface};
 
 mod support;
--- a/examples/deferred.rs
+++ b/examples/deferred.rs
@@ -4,6 +4,7 @@ extern crate cgmath;
 extern crate image;
 
 use glium::index::PrimitiveType;
+#[allow(unused_imports)]
 use glium::{glutin, Surface};
 use std::io::Cursor;
 
--- a/examples/displacement_mapping.rs
+++ b/examples/displacement_mapping.rs
@@ -3,6 +3,7 @@ extern crate image;
 extern crate glium;
 
 use std::io::Cursor;
+#[allow(unused_imports)]
 use glium::{glutin, Surface};
 
 mod support;
--- a/examples/fullscreen.rs
+++ b/examples/fullscreen.rs
@@ -7,6 +7,7 @@ use std::io::Cursor;
 
 use glium::Surface;
 use glium::index::PrimitiveType;
+#[allow(unused_imports)]
 use glium::glutin::{self, ElementState, VirtualKeyCode, Event, WindowEvent};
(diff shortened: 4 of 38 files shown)


est31:
Ok found another little warning (only visible when doing cargo test). They are all gone now, except for one fixme.

ZeGentzy:
Since we are fixing warnings, how about these:
```
  15:16  14.04.19  cargo run --example tutorial-13
    Updating crates.io index
   Compiling syn v0.15.31
   Compiling glutin v0.20.0
   Compiling glium v0.24.0 (/home/gentz/Documents/gfx/glium)
warning: unused doc comment
  --> src/program/mod.rs:58:1
   |
58 | / /// Some shader compilers have race-condition issues, so we lock this mutex
59 | | /// in the GL thread every time we compile a shader or link a program.
   | |______________________________________________________________________^
60 |   // TODO: replace by a StaticMutex
61 | / lazy_static! {
62 | |     static ref COMPILER_GLOBAL_LOCK: Mutex<()> = Mutex::new(());
63 | | }
   | |_- rustdoc does not generate documentation for macro expansions
   |
   = note: #[warn(unused_doc_comments)] on by default
   = help: to document an item produced by a macro, the macro must produce the documentation as part of its expansion

   Compiling num-derive v0.2.4
   Compiling tiff v0.2.2
   Compiling image v0.21.1
warning: unused `std::result::Result` that must be used
   --> src/context/mod.rs:530:9
    |
530 | /         ops::read(&mut ctxt, ops::Source::DefaultFramebuffer(gl::FRONT_LEFT), &rect,
531 | |                           &mut data, false);
    | |____________________________________________^
    |
    = note: #[warn(unused_must_use)] on by default
    = note: this `Result` may be an `Err` variant, which should be handled

warning: the item `glutin` is imported redundantly
 --> examples/tutorial-13.rs:8:17
  |
8 |     use glium::{glutin, Surface};
  |                 ^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

    Finished dev [unoptimized + debuginfo] target(s) in 56.37s
     Running `target/debug/examples/tutorial-13`

```

est31:
Hmmm seems these warnings are spawned by rustc nightly  now...

> unused doc comment

This one is easily fixable.

> unused `std::result::Result` that must be used

This needs a breaking API change, but we already do one with #1739 so that's okay.

> the item `glutin` is imported redundantly

For this one, I think the warning is at fault. It was added only recently: https://github.com/rust-lang/rust/pull/58805. One could blindly fix it by removing the use, but the reason there's an use in the first place and not an extern crate is because the example is supposed to be self-contained. i.e. you should be able to copy it into a separate crate and it should run. If the use is there, it works, if it isn't you have to add glutin to the Cargo.toml but it's redundant as glium already reexports glutin. So i guess I'll add a bunch of `allow`s.

est31:
@mitchmindtree I think this can be merged now!

PR #1743, opened 2019-05-08
Allow instancing for OpenGL ES from version 3.0

0ax1:
Instancing was introduced for OpenGL ES in 3.0. This PR extends the check if instancing is supported for the given OpenGL ES context and extracts the check into a separate function.


0ax1:
Closing this, realizing the level of maintenance is low at this point on Glium.

PR #1757, opened 2019-07-23
Updating to glutin 0.22.0-alpha1 / winit 0.22.0-alpha1

KeyMaster-:
Updated the dependencies, and changed the parts of the code where the EventLoop struct appears. This bumps the glium version to 0.25.1-alpha1.
The examples have not been updated or tested yet.


est31:
Thanks for the PR! Can you change the version to `0.26.0-alpha1`? That's what I meant originally but I wrote 0.25.0-alpha1... my fault :p.

est31:
We can push the alpha release out without fixing the examples, as it seems to be a lot of work and @phaazon really wants an alpha release. But I'd certainly *prefer* the examples to be fixed.

KeyMaster-:
Oh I see, the icon loading complication occurs in the examples. I'll updated the version number first, and then I'll have a look at how much work fixing the examples is.

est31:
Hmm also I wished to merge with the examples failing but it seems I can't. Also not sure whether it's uploading to crates.io if travis CI check fails... so I recommend fixing the examples first.

KeyMaster-:
Just pushed changes that fix a good chunk of the examples. The icon loading feature seemed to actually never be used, so I just got rid of that.
I couldn't fix all the examples yet, since winit's structure has changed, and `poll_events` doesn't exist anymore. My current solution is a stopgap, properly implementing this needs changes to how the game loop is handled in examples in general.
I wrote up a little more detail on that, and which examples are working/not working in the [commit message](https://github.com/KeyMaster-/glium/commit/84f82d3098fbc75aa22160b47bf0a2bdada01f07). The only one that runs but may not be working is displacement_mapping, so I'd appreciate if someone else can also test that one.

PR #1542, opened 2016-11-24
Fix picking example

casey:
Includes @Mufferaw's fix for the camera position, as well as using saturated sub to avoid a panic when the mouse is at the bottom of the screen.

PR #1763, opened 2019-07-24
Fix camera position and use saturating subtraction to get bottom.

casey:
Includes @Mufferaw's fix for the camera position, as well as using saturated sub to avoid a panic when the mouse is at the bottom of the screen.

(I opened a new PR, since I had deleted my fork since opening it, so I couldn't update #1542.)


casey:
No worries, and thank you for merging!

PR #1777, opened 2019-08-17
Stop using mem::transmute(0usize) for get_val invocation

### LINKED ISSUE #1776: Undefined Behavior on `implement_vertex!` macro expansion
This is UB:

https://github.com/glium/glium/blob/5a4255e09c3df66367353fab45d5302fb2123cd2/src/macros.rs#L139

  - (A rust reference is **always** non-null)

  - Given that it is only used to extract the type of a value, something like this could be used instead:

    ```rust
    fn attr_type_of_val<T> (_: Option<&T>) -> $crate::vertex::AttributeType
    where
        T : $crate::vertex::Attribute,
    {
        <T as $crate::vertex::Attribute>::get_type()
    }

    let dummy: $struct_name;
    let mb_ref: Option<&_> = if true {
        None
    } else {
        dummy = unsafe { ::std::mem::zeroed() };
        // assert existence of field name without deref coercions
        let $struct_name { $field_name: _, .. } = dummy;
        Some(&dummy.$field_name)
    };
    attr_type_of_val(mb_ref)
    ```

      - solves one of the issues spotted in #1775

est31:
Fixes #1776

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -131,13 +131,12 @@ macro_rules! implement_vertex {
                                 offset
                             },
                             {
-                                fn attr_type_of_val<T: $crate::vertex::Attribute>(_: &T)
+                                fn attr_type_of_val<T: $crate::vertex::Attribute>(_: Option<&T>)
                                     -> $crate::vertex::AttributeType
                                 {
                                     <T as $crate::vertex::Attribute>::get_type()
                                 }
-                                let dummy: &$struct_name = unsafe { ::std::mem::transmute(0usize) };
-                                attr_type_of_val(&dummy.$field_name)
+                                attr_type_of_val(None::<&$struct_name>.map(|v| v.$field_name))
                             },
                             false
                         )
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -131,13 +131,12 @@ macro_rules! implement_vertex {
...
                             {
-                                fn attr_type_of_val<T: $crate::vertex::Attribute>(_: &T)
+                                fn attr_type_of_val<T: $crate::vertex::Attribute>(_: Option<&T>)
                                     -> $crate::vertex::AttributeType
                                 {
                                     <T as $crate::vertex::Attribute>::get_type()
                                 }
-                                let dummy: &$struct_name = unsafe { ::std::mem::transmute(0usize) };
-                                attr_type_of_val(&dummy.$field_name)
+                                attr_type_of_val(None::<&$struct_name>.map(|v| v.$field_name))
```
RalfJung: Whatever is going on here looks clever... might be a good idea to add a comment explaining it.

The `map` closure is dead code, right? Just present to make the type checker do stuff?

danielhenrymantilla: You can look at #1776 for context

danielhenrymantilla:
Good job!

src/macros.rs
RalfJung: Sure, but that doesn't help for someone reading the code in half a year.

src/macros.rs
```diff
@@ -166,13 +165,12 @@ macro_rules! implement_vertex {
...
                             {
-                                fn attr_type_of_val<T: $crate::vertex::Attribute>(_: &T)
+                                fn attr_type_of_val<T: $crate::vertex::Attribute>(_: Option<&T>)
                                     -> $crate::vertex::AttributeType
                                 {
                                     <T as $crate::vertex::Attribute>::get_type()
                                 }
-                                let dummy: &$struct_name = unsafe { ::std::mem::transmute(0usize) };
-                                attr_type_of_val(&dummy.$field_name)
+                                None::<&$struct_name>.map(|v| v.$field_name)
```
est31: Oh no this is wrong...

src/macros.rs
est31: It's not *that* clever to me. It's a method to obtain the type of the `$field_name` field and call get_type on it without having to construct an instance of the struct.

est31: Also it's safe code now. But I'll add a comment regardless.

PR #1778, opened 2019-08-17
Add a comment explaining what we do in the implement_vertex macro

est31:
Also add a forgotten attr_type_of_val invocation.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -138,7 +138,7 @@ macro_rules! implement_vertex {
                                 {
                                     <T as $crate::vertex::Attribute>::get_type()
                                 }
-                                let field_option = None::<&$struct_name>.map(|v| v.$field_name);
+                                let field_option = None::<&$struct_name>.map(|v| &v.$field_name);
                                 attr_type_of_val(field_option)
                             },
                             false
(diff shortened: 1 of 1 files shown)


src/macros.rs
```diff
@@ -131,12 +131,15 @@ macro_rules! implement_vertex {
...
                             {
+                                // Obtain the type of the $field_name field of $struct_name and
+                                // call get_type on it.
                                 fn attr_type_of_val<T: $crate::vertex::Attribute>(_: Option<&T>)
                                     -> $crate::vertex::AttributeType
                                 {
                                     <T as $crate::vertex::Attribute>::get_type()
                                 }
-                                attr_type_of_val(None::<&$struct_name>.map(|v| v.$field_name))
+                                let field_option = None::<&$struct_name>.map(|v| v.$field_name);
```
RalfJung: Why isn't this `&v.$field_name`?

RalfJung: How did this even compile before?^^

est31: The macro itself compiled but expansion only happens in the tests which right now always fail on travis. So I didn't notice that the test for `implement_vertex!` failed as well.

PR #1779, opened 2019-08-18
Fix unsoundness in implement_vertex

thomcc:
Creating a `&T` with a zero value is never valid. In principal, it might not be valid to create a std::mem::zeroed T but it's hard for me to imagine it goes wrong for any real-world implementation of Vertex (and if it did things like transform feedback would be unsound as well).

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -162,9 +162,11 @@ macro_rules! implement_vertex {
                         (
                             Cow::Borrowed(stringify!($field_name)),
                             {
-                                let dummy: &$struct_name = unsafe { ::std::mem::transmute(0usize) };
-                                let dummy_field = &dummy.$field_name;
-                                let dummy_field: usize = unsafe { ::std::mem::transmute(dummy_field) };
+                                let dummy: $struct_name = unsafe { ::std::mem::zeroed() };
+                                let dummy_addr = &dummy as *const $struct_name as usize;
+                                let dummy_offset = &dummy.$field_name as *const _ as usize;
+                                let dummy_field: usize = dummy_offset - dummy_addr;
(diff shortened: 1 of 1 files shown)


thomcc:
I think any implementation of Vertex for which this is unsound is fundamentally unsound to use in several cases (transform feedback, being the obvious one -- if your struct has fields with invalid bitpatterns, the GPU cannot know about these, and will not respect them!)

But if you'd rather this be done with memoffset, fine.

est31:
Yeah so actually when 0's are a valid bitpattern for the vertex, then offset_of is in fact even harmful because internally it uses uninitialized memory which is UB.

thomcc:
Another solution would just be to require Default::default() on Vertex. It's a breaking change but not one that should be hard for anyone to adopt. Doing so would remove the need for unsafety.

thomcc:
I also must admit I'm unsure how to reexport the memoffset::offset_of macro -- the details of macro importing and exporting have always been a bit fuzzy to me, especially in pre-2018 Rust.

est31:
> Another solution would just be to require Default::default() on Vertex.

This is a quite compelling solution as it would completely avoid unsound code. The codebase with most needed changes due to this would probably be us as glium uses `implement_vertex` quite extensively in its own examples tests. Optimally, I'd love to see the `implement_vertex!` macro gone completely and replaced with a derive macro (like readme also mentions). With that breaking change, we could require Default to be implemented as well.

> I also must admit I'm unsure how to reexport the memoffset::offset_of macro

Okay, let me do a PR for this myself. I already wanted to do it yesterday.

thomcc:
All yours!

est31:
@thomcc see #1782

Issue #1565, opened 2017-01-27
Cannot implement a transform feedback shader that receives and writes to a VertexBuffer with same vertex type


ColonelThirtyTwo:
I can't figure out how to write and use a transform feedback enabled vertex shader that takes in as well as writes to a `VertexBuffer` of the same type.

The crux of the issue is that glium matches shader vertex attribute variables as well as transform feedback output variables with vertex buffer attributes based on their names; i.e. for a simple vertex structure `struct Vertex { position: (f32, f32, f32, f32) }`, `position` would be matched to the shader vertex attribute variable `in vec4 position;`

This is an issue when using transform feedback, as for glium to properly bind the shader variables correctly, the shader must have an `in vec4 position;` variable for the vertex attribute coming in, as well as a `out vec4 position;` variable for the transform feedback outputs. This is obviously impossible, since it creates a name conflict in the shader. And the developer cannot rename either of the variables, since glium will not be able correctly bind them.

What can I do to correct this issue, and use transform feedback where both the input and the output buffers have the same type?

My best guess would be to create another vertex type that has the exact same structure as the real vertex type, but has different field names, and unsafely cast the `VertexBuffer` to the fake type, effectively renaming the fields, but I can't figure out how to do that.

(My use case is a double-buffered particle system; the transform feedback shader would read a vertex buffer containing the current particle states, do a step of the simulation, and write the new particle state to another vertex buffer, which would be read by the same shader the next frame.)

unknownue:
The same problem occurs to me when I try to implement a particle example using transform feedback feature. This problem does make transform feedback hard to use. I spent hours to find solution to it and finally gave up.

PR #1788, opened 2019-09-07
Remove the check of naming when buiding TransformFeedbackSession.

unknownue:
When creating `TransformFeedbackSession`, `glium` would check the consistent of naming, offset and type between the input attributes of output buffer and output attributes of input buffer(both are attributes in vertex buffer).
The offset and type checking is OK. But the check of naming make TransformFeedback difficult to use.

Because when implementing particles system, we usually use two transformfeedback buffer(`buffer1` and `buffer2`) with the same vertex type. In first render pass, `buffer1` is for input and `buffer2` is for output. Then in the second render pass, we swap the position of them(e.g. `buffer1` is for output, `buffer2` is for input).
The problem is that `glium` will stop me due to a fact that the naming of input and output attributes are different. But they must be different, or we can't distinguish what is input attributes and output attributes. This causes a self-contradiction condition.

So we just need to remove the check for naming in TransformFeedback buffer and then everything works fine.

### DIFF
--- a/src/program/raw.rs
+++ b/src/program/raw.rs
@@ -405,6 +405,10 @@ impl RawProgram {
     /// and `stride`.
     ///
     /// The `stride` is the number of bytes between two vertices.
+    ///
+    /// This function only check the identity for type and offset(exclude the naming) between vertex attributes.
+    ///
+    /// The correctness of semantic meaning(i.e. naming) between vertex attributes is the responsibility of user.
     pub fn transform_feedback_matches(&self, format: &VertexFormat, stride: usize) -> bool {
         // TODO: doesn't support multiple buffers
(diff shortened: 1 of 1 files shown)


unknownue:
Fix #1565.

unknownue:
Is this what you mean?
```
for elem in buf.elements.iter() {
    if format.iter().find(|e| e.1 == elem.offset && e.2 == elem.ty)
                    .is_none()
    {
        return false;
    }
    
    if format.iter().any(|e| e.1 != elem.offset && e.0 == elem.name) {
        return false;
    }
}
```

est31:
@unknownue yes, that'd be very useful.

Issue #1780, opened 2019-08-18
segfaults since the glutin update


est31:
Since the glutin/winit update PR #1774, I've observed segfaults when cargo test is ran. They are also visible on travis CI. They are of forms like:

```
     Running `/home/travis/build/glium/glium/target/debug/deps/attrs-ca220506cc1fbab9`

running 51 tests

Xlib: sequence lost (0x10080 > 0x84) in reply type 0x0!

Xlib: sequence lost (0x1008a > 0x8c) in reply type 0x0!

Xlib: sequence lost (0x1008f > 0x9a) in reply type 0x0!

error: process didn't exit successfully: `/home/travis/build/glium/glium/target/debug/deps/attrs-ca220506cc1fbab9` (signal: 11, SIGSEGV: invalid memory reference)
```

They exist both on travis CI as well as on my own computer. When you try to run the attrs test binary manually or any subset of the individual tests, the segfault doesn't happen. They also happen with other tests than the attrs one, and occur on all rustc versions I've tried: 1.36.0, 1.37.0, and nightly.

As my glutin update PR only changed the glutin version, I'm not sure whether the cause of the bug lies in glium or whether it lies in upstream libraries instead. cc @ZeGentzy @osspial @francesca64 on this.

ZeGentzy:
Backtrace in gdb? The SEGV and sequence lost errors are I suspect for different reasons.

est31:
Yes the sequence lost errors are likely unrelated, they also occur when there is no segfault.

I've been trying to get a GDB backtrace but I failed. The segfaults only occur when you do cargo test and then in a child so you need to tell gdb to follow forks but for some reason I'm too stupid for that. I will try again in an hour or so after finishing some other work.

ZeGentzy:
I think you should be able to do the following:
```
set detach-on-fork off
info inferiors
inferiors <proccess you want>
```

est31:
Yeey I've got a backtrace:

<details>

```
#0  0x00007ffff74b1fd3 in _XUnregisterFilter () from /usr/lib/x86_64-linux-gnu/libX11.so.6
#1  0x00007ffff74df9b7 in ?? () from /usr/lib/x86_64-linux-gnu/libX11.so.6
#2  0x00007ffff74c5b32 in XDestroyIC () from /usr/lib/x86_64-linux-gnu/libX11.so.6
[...]

PR #1814, opened 2020-01-06
Report which shader stage failed to compile

avl:
When compiling a program composed of multiple shader stages, glium would not previously report which shader stage failed to compile. The error message from the driver would include line numbers, but the user had to guess which file actually failed to compile.

This commit makes the error data type include which shader stage actually failed to compile.

### DIFF
--- a/src/program/mod.rs
+++ b/src/program/mod.rs
@@ -62,11 +62,57 @@ lazy_static! {
     static ref COMPILER_GLOBAL_LOCK: Mutex<()> = Mutex::new(());
 }
 
+/// Used in ProgramCreationError::CompilationError to explain which shader stage failed compilation 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ShaderType {
+    /// Vertex shader, maps to gl::VERTEX_SHADER
+    Vertex,
+    /// Geometry shader, maps to gl::GEOMETRY_SHADER
+    Geometry,
+    /// Fragment shader, maps to gl::FRAGMENT_SHADER
+    Fragment,
+    /// Tesselation control shader, maps to gl::TESS_CONTROL_SHADER
+    TesselationControl,
+    /// Tesselation evaluation shader, maps to gl::TESS_EVALUATION_SHADER
+    TesselationEvaluation,
+    /// Compute shader, maps to gl::COMPUTE_SHADER
+    Compute,
+}
+
+impl ShaderType {
+    /// Creates an instance of gl::types::GLenum corresponding to the given ShaderType
+    pub fn to_opengl_type(self) -> gl::types::GLenum {
+        match self {
+            ShaderType::Vertex => gl::VERTEX_SHADER,
+            ShaderType::Geometry => gl::GEOMETRY_SHADER,
+            ShaderType::Fragment => gl::FRAGMENT_SHADER,
+            ShaderType::TesselationControl => gl::TESS_CONTROL_SHADER,
+            ShaderType::TesselationEvaluation => gl::TESS_EVALUATION_SHADER,
+            ShaderType::Compute => gl::COMPUTE_SHADER,
+        }
+    }
+    /// Creates an instance of ShaderType corresponding to the given gl::types::GLenum.
+    /// This routine will panic if the given shadertype is not supported by glium.
+    pub fn from_opengl_type(gl_type: gl::types::GLenum) -> Self {
+        match gl_type {
+            gl::VERTEX_SHADER => ShaderType::Vertex,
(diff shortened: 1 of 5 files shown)


avl:
Note that this is technically an API-breaking change, since someone might be matching or destructuring the 'ProgramCreationError'-enum, and this PR changes the member 'CompilationError'.

est31:
> Note that this is technically an API-breaking change

That's fine! The next release will be an API-breaking one!

src/program/mod.rs
```diff
@@ -62,11 +62,57 @@ lazy_static! {
     static ref COMPILER_GLOBAL_LOCK: Mutex<()> = Mutex::new(());
 }
 
+/// Used in ProgramCreationError::CompilationError to explain which shader stage failed compilation 
+#[derive(Copy, Clone, Debug)]
```
est31: Maybe also derive Eq?

src/program/mod.rs
```diff
@@ -109,8 +155,16 @@ impl Error for ProgramCreationError {
...
-            CompilationError(_) =>
-                "Compilation error in one of the shaders",
+            CompilationError(_,typ) => {
+                match typ {
+                    ShaderType::Vertex => "Compilation error in vertex shader",
+                    ShaderType::Geometry => "Compilation error in geometry shader",
+                    ShaderType::Fragment => "Compilation error in fragment shader",
+                    ShaderType::TesselationControl => "Compilation error in tesselation control shader",
+                    ShaderType::TesselationEvaluation => "Compilation error in tesselation evaluation shader",
+                    ShaderType::Compute => "Compilation error in one of the compute shader"
```
est31: compute shaders*

src/program/mod.rs
avl: I'll fix.

src/program/mod.rs
avl: Good catch!

src/program/mod.rs
avl: I'll also make sure to rebase and get rid of the merge commit. I had attempted to do so, but obviously made some sort of mistake.

src/program/mod.rs
```diff
@@ -109,8 +155,16 @@ impl Error for ProgramCreationError {
...
-            CompilationError(_) =>
-                "Compilation error in one of the shaders",
+            CompilationError(_,typ) => {
+                match typ {
+                    ShaderType::Vertex => "Compilation error in vertex shader",
+                    ShaderType::Geometry => "Compilation error in geometry shader",
+                    ShaderType::Fragment => "Compilation error in fragment shader",
+                    ShaderType::TesselationControl => "Compilation error in tesselation control shader",
+                    ShaderType::TesselationEvaluation => "Compilation error in tesselation evaluation shader",
+                    ShaderType::Compute => "Compilation error in compute shader"
```
est31: "s" is still missing

src/program/mod.rs
avl: I reworded the string.

avl:
I just realized I forgot to address the winit-issue @est31 mentioned. Looking into it now.

est31:
@avl no it's alright. All I asked was a rebase which you did. I'm waiting for CI to pass and then I'll merge the PR.

avl:
Oops, I see I should have run the test suite. I'll fix the test failure.

avl:
I fixed the compilation failure in the tests, but my machine all tests seem to with our without the changes in this PR. Is this a known issue?

The error I get is:

`
---- attribute_float_f32 stdout ----
thread 'attribute_float_f32' panicked at 'Initializing the event loop outside of the main thread is a significant cross-platform compatibility hazard. If you really, absolutely need to create an EventLoop on a different thread, please use the 'EventLoopExtUnix::new_any_thread' function.', /home/anders/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.20.0-alpha4/src/platform_impl/linux/mod.rs:698:9
`

avl:
Ah, wait a minute, _that_ is of course the test failure you were referring to?

est31:
> Is this a known issue?

Yes! It's basically the same issue as #1780

> Ah, wait a minute, that is of course the winit issue you were referring to?

No that issue would manifest in build failure of glium... the problem is that winit 0.20.0 was released but old glutin alphas depended alpha releases of it and there was a break in the winit API between the alpha releases and the official 0.20.0 release, but if you depended on the old glutin alphas, cargo would still combine them with the official release, leading to build failure.... The fix had to be done in glutin to hardcode the winit release, and we had to update glutin, which was done in #1815 .

PR #1815, opened 2020-01-06
Use glutin 0.22.0-alpha6

2xsaiko:
0.22.0-alpha6 locks winit to a compatible version, it fails to compile otherwise because it tries to use winit 0.20.0 instead of 0.20.0-alpha4 (see https://github.com/rust-windowing/glutin/commit/5947a22bbd62fe3b39601d6100563aaab352a315#diff-1b6d60f3f005be6e1e05f2803da7ccc6)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ unstable = [] # used for benchmarks
 test_headless = []  # used for testing headless display
 
 [dependencies.glutin]
-version = "=0.22.0-alpha5"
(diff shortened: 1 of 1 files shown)


est31:
Oh yeah that locking is important indeed. As you can see we did it for glium as well due to similar issues. I've already wanted to suggest to the glutin maintainers to lock winit as well but seems that they beat me to it. Wonderful!

PR #1829, opened 2020-02-29
Fix documentation tests.

baileyn:
Fixes the documentation tests so that `cargo test --doc` will succeed.

### DIFF
--- a/src/field.rs
+++ b/src/field.rs
@@ -23,7 +23,7 @@ impl<T> Field<T> {
 }
 
 #[doc(hidden)]
-pub fn _hidden_field<T>(offs: usize, size: usize, _: Option<&T>) -> Field<T> {
+pub fn _hidden_field<T>(offs: usize, _: usize, _: Option<&T>) -> Field<T> {
     Field {
         offs,
         size: std::mem::size_of::<T>(),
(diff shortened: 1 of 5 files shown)


src/field.rs
```diff
@@ -37,6 +37,6 @@ macro_rules! field {
     ($struct_name:ident, $field_name:ident) => {{
         let opt = None::<&$struct_name>.map(|v| &v.$field_name);
         let offs = $crate::__glium_offset_of!($struct_name, $field_name);
-        $crate::field::_hidden_field(offs, size, opt)
```
est31: Could you maybe remove the size param entirely? It's not needed for anything.

baileyn: I've went ahead and removed the size parameter :)

est31:
Could you do a git rebase instead of merging my commits? The diff is messed up with the changes from the master branch.

baileyn:
@est31 they're going through the pipeline now. Out of curiosity, does it always require a force push or am I doing something wrong?

est31:
@baileyn yes a rebase always needs a force push. That's expected and okay. The goal is to have a clean history, which merge commits inside PRs don't give you.

PR #1832, opened 2020-03-08
resolve correct paths in glutin 0.23.0

ikamensh:
This fixes paths to the structs needed in the tutorial. Note that tutorial still doesn't work with these fixes due to 
```
error[E0599]: no method named `poll_events` found for type `glium::glutin::event_loop::EventLoop<()>`
```


ikamensh:
I was able to create a PR with code that compiles https://github.com/glium/glium/pull/1833, closing this one.

Issue #1827, opened 2020-02-23
Implicit naming convention 


InnocentusLime:
I haven't found this being mentioned in the docs for `implement_vertex` macros, but apparently the names of the fields of the struct you are using for a VBO must match the attribute names in the shader or else glium will panic (not return an error code!) claiming that an attribute "is missing in vertex array bindings".

est31:
PR to point this out in the `implement_vertex` docs is welcome!

InnocentusLime:
Okay. I'll create a PR for that! That should be simple... Just a pair of new lines. However, I am a bit confused why Glium prefers panicing with a scary message rather then generating an error code.

InnocentusLime:
Created a PR for that. The id is #1841

est31:
Fixed by #1841

PR #1841, opened 2020-04-23
Attempt to solve issue #1827

InnocentusLime:
I added some information to docs, because it seemed like there was some undocumented requirements for structs which are fed to the `implement_vertex` macro which must be mentioned.


est31:
So it's been successful on travis CI but for some reason github couldn't get the status, and without github registering the success I can't merge the PR. I'm not sure how to resolve it.

I'll close then reopen maybe that helps.

PR #1843, opened 2020-04-26
Add small note that glium uses column-major matrices

BenSandeen:
This simple paragraph would've saved me about half an hour, so I thought I'd add it for future readers

### DIFF
--- a/book/tuto-04-matrices.md
+++ b/book/tuto-04-matrices.md
@@ -47,7 +47,7 @@ target.draw(&vertex_buffer, &indices, &program, &uniforms,
             &Default::default()).unwrap();
 ```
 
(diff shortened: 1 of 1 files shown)


book/tuto-04-matrices.md
```diff
@@ -47,6 +47,14 @@ target.draw(&vertex_buffer, &indices, &program, &uniforms,
             &Default::default()).unwrap();
 ```
 
+Note that in glium the matrices are column-major.  If we were to write the above matrix in standard mathematical notation, it would look like this (unfortunately, markdown does not let us format things with LaTeX, so this doesn't look nicely formatted):
```
est31, line 50: Could you write "in glium and opengl"? Because all of opengl assumes column major matrices.

est31: Also the comment in the () isn't needed I think.

BenSandeen: Done!

Issue #1840, opened 2020-04-16
gl function was not loaded


virtualritz:
I'm seeing [this issue](https://github.com/glium/glium/issues/1648) with glium 0.27.0 on macOS 10.15.4 on a 2014 MBP with Intel Iris GPU.
```
thread 'main' panicked at 'gl function was not loaded', /Users/moritz/code/generative/target/release/build/glium-5a230c4287d7cff8/out/gl_bindings.rs:2535:13
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::panicking::default_hook::{{closure}}
   4: std::panicking::default_hook
   5: std::panicking::rust_panic_with_hook
   6: std::panicking::begin_panic
   7: glium::gl::missing_fn_panic
   8: glium::texture::any::new_texture
   9: valora::gpu::Gpu::build_texture
  10: valora::run_fn
  11: generative::main
  12: std::rt::lang_start::{{closure}}
  13: std::panicking::try::do_call
  14: __rust_maybe_catch_panic
  15: std::rt::lang_start_internal
  16: main
```

chengluyu:
Me, too. I was also trying [valora](https://paytonturnage.gitbook.io/valora/) and met this error on macOS 10.15.4

est31:
I don't have a Mac computer. I wonder how a minimal reproducing example could look like.

virtualritz:
I'm using this example from the `valora` guide:
```Rust
use valora::prelude::*;

fn main() -> Result<()> {
    run_fn(Options::from_args(), |_gpu, world, _rng| {
        Ok(move |ctx: Context, canvas: &mut Canvas| {
            canvas.set_color(LinSrgb::new(1., 1., 1.));
            canvas.paint(Filled(ctx.world));

            let max_radius = world.width / 3.;
            let radius =
                ctx.time.as_secs_f32().cos().abs() * max_radius;

            canvas.set_color(LinSrgb::new(1., 0., 0.));
            canvas
                .paint(Filled(Ellipse::circle(world.center(), radius)));
        })
    })
}
```

est31:
Can you reproduce it only with glium's API? And on other operating systems than Mac?

jarhill0:
> And on other operating systems than Mac?

As someone else who has experienced this panic, I think it's a macOS-specific issue.

mlakewood:
[...]

PR #1846, opened 2020-05-17
Bad fix for https://github.com/glium/glium/issues/1840

mlakewood:
This PR fixes https://github.com/glium/glium/issues/1840 on my osx but
    is obviously going to break something else. I dont know enough about
    glium or any of the problems to see if this is appropriate, but I do
    have a mac and can verify that this works now on it. Hopefully somebody
    with a bit more knowledge than me can submit a good fix that works.


est31:
Thanks for the demo, but as this is no serious pull request, I'm closing it.

Issue #1728, opened 2019-02-06
How to enable GL_TEXTURE_CUBE_MAP_SEAMLESS


krupitskas:
So am I right that we can't enable extensions directly with glium? How can i call glEnable for example?

Boscop:
Yes, I also need to enable `GL_TEXTURE_CUBE_MAP_SEAMLESS`. Is there a way now?

Issue #644, opened 2015-03-13
Add support for loading and binding cube map textures


ryanstew:
Ideally including proper handling of GL_TEXTURE_CUBE_MAP_SEAMLESS/ARB_seamless_cube_map.

tomaka:
With #1169 it is now possible to create empty cubemap and cubemap arrays, and bind them to uniforms and to framebuffers. However uploading data isn't supported yet.

wyozi:
So does this mean that uploading data to cubemap currently requires directly accessing opengl?

tomaka:
I'm not sure what you mean. Currently you have to use for example a blit to write data.

wyozi:
Creating a new texture2d with image data is as simple as ```Texture2d::new(display, image)```, but there does not seem to be a similar method to easily upload cubemap faces. I looked at blitting example but Cubemap does not seem to have as_surface so I don't know how blitting would work for it.

tomaka:
You can do `let target = glium::framebuffer::SimpleFrameBuffer::new(&display, cubemap.main_level().image(glium::texture::CubeLayer::PositiveX)).unwrap();` to obtain the same thing as calling `as_surface()` gives you.

wyozi:
Ahh, that works, thanks.

porky11:
There is'nt a way to initialize cube maps by just adding images to it, is it?

Boscop:
Btw, this works: http://onagat.hatenablog.com/entry/2017/03/24/235635

Boscop:
@ryanstew I just added `GL_TEXTURE_CUBE_MAP_SEAMLESS`/`ARB_seamless_cube_map` :)

Boscop:
@tomaka Is it possible to sample a single `Cubemap` face as if it was a `Texture2d`, so that the shader can take it in as `sampler2d`? 
I want to post-process a cubemap's faces with shaders that take textures as input and write to textures. I already have the code for writing the invidual cubemap faces, now I just need to figure out how to sample them as `sampler2d`. Seems like `cubemap.main_level().image(cube_layer)` can't be sampled (?)

0xFlugel:
I was looking at whether I want to implement the creation of a cubemap texture with data, like texture2d etc. have. I don't know much about them though. I may not be the right person for that, though, if there are non-obvious things to consider. Some feedback would be nice.

[...]

PR #1851, opened 2020-06-07
Enable GL_TEXTURE_CUBE_MAP_SEAMLESS when available

Boscop:
Fixes https://github.com/glium/glium/issues/1728 (and maybe https://github.com/glium/glium/issues/644).

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glEnable.xhtml
> GL_TEXTURE_CUBE_MAP_SEAMLESS is available only if the GL version is 3.2 or greater.

---

(Not sure if I really have to add `"GL_ARB_seamless_cube_map"` to the `Registry` in `build/main.rs`. It also compiles without that.)

### DIFF
--- a/build/main.rs
+++ b/build/main.rs
@@ -49,6 +49,7 @@ fn generate_gl_bindings<W>(dest: &mut W) where W: Write {
             "GL_ARB_occlusion_query",
             "GL_ARB_pixel_buffer_object",
             "GL_ARB_robustness",
+            "GL_ARB_seamless_cube_map",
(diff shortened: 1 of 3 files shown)


est31:
I don't know whether it should be enabled unconditionally or not. Would people ever want to turn it off?

Boscop:
@est31 I can't imagine a use case where someone would want it disabled.
And enabling doesn't hurt performance either.

est31:
@Boscop alright then.

PR #1852, opened 2020-06-09
Use consistent explicit type

ducthotran2010:


est31:
Thanks for the PR but I'll close this. From the `[]` you can see that this is an array constructor, and array content always has the same type. So the single f32 here controls the scalar type of the entire matrix. It makes the code clearer if the type is explicit so I'd prefer to keep it. Thanks again.

Issue #1767, opened 2019-08-03
Get rid of mem::uninitialized completely


est31:
Continuation of #1753. @Lokathor has removed mem::uninitialized usage in the glium crate itself in #1759, but some examples still contain it.

LiamTheProgrammer:
I think that https://github.com/glium/glium/pull/1853 could solve this? I used ripgrep and sed to replace all instances of it.

est31:
#1853 replaced uninitialized with assume_init, but optimally we'd have functions as described [here](https://github.com/glium/glium/pull/1853#issuecomment-644941195).

Lokathor:
Just glancing at the changes in #1853, it's _almost certainly_ wrong and UB, but they're `no_run` examples so... maybe it doesn't matter.

Then again, examples should probably be correct because someone probably copies them into their project as a starting point.

PR #1853, opened 2020-06-16
Replace all instances of mem::uninitialized in comments with assume_init

LiamTheProgrammer:
Since mem::uninitialized is a deprecated part of std, I figured that through some search and replace the comments could be improved. The instances of mem::uninitialized got replaced with MaybeUninit and assume_init, which is still unsafe, but at least it's not deprecated. I ran all tests on my computer and built glium successfully.


LiamTheProgrammer:
I think this might solve https://github.com/glium/glium/issues/1767

est31:
Doing assume_init without initializing is as ub as using uninitialized. I'm not really sure the new solution is better. Could you instead put the example code into functions like:

```
#fn foo(display: glium::Display) {
display.do_something();
#}
```

est31:
Also, could you please delete the todo.txt file? thanks!

LiamTheProgrammer:
At least it isn't deprecated. But yeah, I'll remove todo.txt. Since the uninitialized memory is handled in the right way here once it's made IIRC, I don't think it's a danger to have it this way.

est31:
Thanks for removing the todo file. Could you also remove the assume_init in favour of the function based approach I described?

LiamTheProgrammer:
@est31 Ok.

LiamTheProgrammer:
@est31 I'll look back at this tomorrow.

est31:
@LiamTheProgrammer deal!

PR #1854, opened 2020-06-17
Add lots of documentation and remove #[allow(missing_docs)] from lots of files

LiamTheProgrammer:
This, if merged, would add significant amounts of documentation for previously undocumented items. It would also completely remove `#[allow(missing_docs)]` from a lot of files. Also, it clarifies that support for compute shaders does exist in OpenGL by fixing one `FIXME` comment.

The documentation in lots of places isn't ideal, but from what I understand it's accurate. This is a somewhat significant step towards having full and in-depth documentation.


est31:
Thanks for the PR! It looks mostly good so far, but regarding order of things I'd prefer a PR that removes assume_init first. After that one is merged I'll review this one.

0xFlugel:
I had a short look into these commits. Some reference links e.g. for the `BufferType` seem helpful to someone new to OpenGL. Much of the rest, however, doesn't look like a valuable addition.

Take a look at [this one](https://github.com/glium/glium/pull/1854/commits/cbc3ce8ee03368a96c52a2d0052fbe9405a05506#diff-eb01c2c7b6bfb9f2e174ad610c0a85cf) for example. Nothing is gained from this type of doc strings. Rather, the valuable (already existing) lines with notes about limited support (s. the additions to `UncompressedIntFormat`) are now more easily overlooked. Things like the basic and easy to guess meaning (e.g. of character-number-combinations for internal buffer types) should (at best) be a single, short explanation in the encompassing documentation because it is an overarching concept. It shouldn't be _n_ lines of mostly identical text.

est31:
Yeah the `/*do foo*/ foo()` type doc comments are part of why I'm so reluctant about merging this. Documentation improvements in general are welcome, but I agree that such class of changes don't constitute as improvements. Thus I'll close this for now.

Issue #1876, opened 2020-10-04
release 0.28


aentity:
Release 0.28, new winit 0.23

est31:
#1874 should have done the 0.28 release, but sadly the crates.io token used for publishing glium releases doesn't work any more, because recently crates.io has made all old tokens invalid because of a security issue with token generation.

I've contacted @tomaka (the only person who can set up such a token) via e mail and discord but haven't gotten a response since.

est31:
#1877 succeeded, and 0.28 is out. Thanks everyone.

PR #1877, opened 2020-10-05
Update the token for travis publication

tomaka:
I've run `travis encrypt CRATESIO_TOKEN=<mytoken>` and it gave me `Cw+Zq...`.
I'm unfortunately not sure whether the previous `CRATESIO_TOKEN` was `Ts/4pe...` or `Afe+PL...`, so I'm guessing the latter, but it might be the former.

cc https://github.com/glium/glium/issues/1876


tomaka:
Looking at the travis build of this branch, I think I got it right.

est31:
Let's just merge this and if it pushes the 0.28 release to crates.io, we know it works :).

est31:
Yaay it worked! https://crates.io/crates/glium/0.28.0

Thanks @tomaka for your help. Btw help would be appreciated for rodio as well: https://github.com/RustAudio/rodio/issues/321

PR #1886, opened 2020-10-23
Update Rust edition

agluszak:
This PR changes the edition of Rust used by glium to 2018. It consists almost exclusively of auto-fixes generated by `cargo fix` and `clippy`.  This is my first contribution to a Rust crate :)

Why? I started learning both Rust and OpenGL. When I was following the tutorial from [the book](https://github.com/glium/glium/tree/master/book), I noticed that it was slightly out of date. So I started updating it. I will submit a separate PR with the new version. But then I noticed that the crate itself uses 2015 edition and I thought: "Oh, maybe I could do something with that as well" :)


examples/support/mod.rs
```diff
@@ -1,6 +1,6 @@
 #![allow(dead_code)]
 
-extern crate obj;
+
```
est31, line 3: Why do you replace the `extern crate` statements with empty lines? If you remove them, please also remove the line.

tests/shaders.rs
```diff
@@ -265,7 +265,7 @@ fn get_program_binary() {
         Ok(bin) => bin
     };
 
-    assert!(binary.content.len() >= 1);
+    assert!(!binary.content.is_empty());
```
est31, line 268: I think the `is_empty` lint of clippy is misguided/harmful. The older code was shorter as well as clearer.

est31 (changes requested):
Will accept the PR if the `is_empty` changes are reverted and the `extern crate` lines are removed instead of being blanked.

est31:
Can you revert all `is_empty` calls? And can you squash the `is_empty` relating commits into the commits that introduced the `is_empty` changes to keep blame intact? Thanks.

agluszak:
Fixed! I didn't notice the blank lines, they were introduced by `cargo fix`

src/context/mod.rs
```diff
@@ -789,7 +783,7 @@ fn check_gl_compatibility(version: &Version, extensions: &ExtensionsList)
         result.push("OpenGL implementation doesn't support blitting framebuffers");
     }
 
-    if result.len() == 0 {
+    if result.is_empty() {
```
est31, line 786: is_empty

src/program/reflection.rs
```diff
@@ -615,7 +615,7 @@ pub unsafe fn reflect_transform_feedback(ctxt: &mut CommandContext, program: Han
         let name = String::from_utf8(name_tmp).unwrap();
 
         if buffer_mode == TransformFeedbackMode::Interleaved {
-            if result.len() == 0 {
+            if result.is_empty() {
```
est31, line 618: is_empty

est31 (changes requested):
I'd love a revert of these changes, right inside the commit that changed them to keep the diff intact. Do you know how to do that? Otherwise check out `git rebase -i`.

agluszak:
I squashed all commits into one, I thought it could be done while merging :) Hopefully it's all okay now

est31 (approved):
Thanks!

Issue #1892, opened 2020-12-11
Migrating to Github Actions


est31:
Open source jobs for Travis CI have been reduced even further. Maybe it's time to think about migration to Github Actions now.

est31:
![Screenshot_20201211_064809](https://user-images.githubusercontent.com/8872119/101868036-d88c7800-3b7c-11eb-86a7-73576d3f7f97.png)

https://www.traviscistatus.com/#month

est31:
Just for reference, a build used to take 45 minutes, and now takes about 4 hours... this isn't bearable any more.

est31:
Alright with  #1894 and #1900 merged, the migration is complete. Thanks @tomaka for the help!

PR #1894, opened 2020-12-11
Replace Travis CI with Github Actions

est31:
Part of #1892


tomaka:
I removed the necessity for the Travis check to pass!

Issue #1890, opened 2020-12-01
Loading SPIR-V shaders as glium `Program`?


Boscop:
Apparently, with opengl 4.6 (or extension `ARB_gl_spirv`) it's possible to load SPIR-V shaders with OpenGL.
Can I just use [`ProgramCreationInput::Binary`](https://docs.rs/glium/0.28.0/glium/program/enum.ProgramCreationInput.html) or how would it be possible with glium? (I have opengl 4.6)

Edit: Ah, it has to use [`GL_SHADER_BINARY_FORMAT_SPIR_V`](https://www.khronos.org/opengl/wiki/SPIR-V).. 
Could SPIR-V be supported in glium by adding a variant `ProgramCreationInput::SpirV` that uses this?

Because I'd really like to load [shaders written in Rust compiled to SPIR-V](https://github.com/embarkstudios/rust-gpu) in glium :)

est31:
PRs welcome!

Boscop:
@est31 In case the opengl version is less than 4.6, the `ARB_gl_spirv` extension would need to be enabled to support this. What's the recommended way to allow code to tell glium to enable an extension?

Boscop:
@est31 I tried to implement loading of SPIR-V shaders here: https://github.com/glium/glium/compare/master...Boscop:spirv-shaders

Unfortunately, when I run the `screenshot-spirv` example, it fails with:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: LinkingError("glLinkProgram triggered GL_INVALID_OPERATION")', examples/screenshot-spirv.rs:38:12

Is it invalid to combine a GLSL vertex shader with a SPIR-V fragment shader?

If someone can help me make progress with this, I'd really appreciate it :)

Boscop:
I fixed it, now it's working :)

The issue was that apparently a GLSL vertex shader cannot be combined with a SPIR-V fragment shader, both must be SPIR-V (or GLSL).
https://www.khronos.org/opengl/wiki/SPIR-V#Linking_SPIR-V

PR #1895, opened 2020-12-12
Add support for loading SPIR-V shaders

Boscop:
This PR adds support for [SPIR-V shader loading](https://www.khronos.org/opengl/wiki/SPIR-V) to glium (with OpenGL >= 4.6 or extension [`ARB_gl_spirv`](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_gl_spirv.txt)).
You can see it in action by running the `spirv` example.

For this I added a new variant `ProgramCreationInput::SpirV`.

Fixes: https://github.com/glium/glium/issues/1890


build/main.rs
```diff
@@ -18,7 +18,17 @@ fn main() {
 }
 
 fn generate_gl_bindings<W>(dest: &mut W) where W: Write {
-    let gl_registry = Registry::new(
+    let gl_registry_4_6 = Registry::new(
```
est31, line 21: Can you explain why creating an entirely new registry is needed? Can't it just be integrated into the gl_registry?

est31 (changes requested):
Looks really great for the most part, only one concern beyond the one I expressed inline: Is there a way to make the program compile the spirv files on the fly? Or at least would it be possible to get both spirv files below 1 kb each? I'm not a fan of adding binary files :).

est31:
When trying to apply this patch locally, I'm getting an error:

```
thread 'main' panicked at 'The program attribute `` is missing in the vertex bindings', src/vertex_array_object.rs:287:17
```

```
   2: glium::vertex_array_object::VertexArrayObject::new
             at ./src/vertex_array_object.rs:287
   3: glium::vertex_array_object::Binder::bind
             at ./src/vertex_array_object.rs:206
   4: glium::ops::draw::draw
             at ./src/ops/draw.rs:156
   5: <glium::Frame as glium::Surface>::draw
             at ./src/lib.rs:1191
   6: screenshot_spirv::main
             at ./examples/screenshot-spirv.rs:55
```

Any ideas? Also would be great to rebase the PR onto latest master, as I can only merge PRs that are in such a state.

build/main.rs
Boscop: Not sure if it's valid to do it in the 4.5 registry.
[This site](https://www.khronos.org/opengl/wiki/SPIR-V) says that it's available in 4.6 core, or with the  `ARB_gl_spirv` extension in earlier versions.
My question for you is: What does the version of a registry mean? Does the version of a registry mean that the extensions specified for it will only be generated when the installed OpenGL version is below the version of the registry?
IOW, how does this translate to `Registry` instantiation:

image

Maybe I should also use [`Profile::Core`](https://docs.rs/gl_generator/0.14.0/gl_generator/enum.Profile.html#variant.Core) instead of `Profile::Compatibility`?

examples/screenshot-spirv.rs
```diff
@@ -0,0 +1,63 @@
...
+    let vertex_buffer = {
+        #[derive(Copy, Clone)]
+        struct Vertex {
+            dummy: f32,
+        }
+
+        implement_vertex!(Vertex, dummy);
+
+        glium::VertexBuffer::<Vertex>::empty(&display, 0).unwrap()
+    };
```
Boscop, line 25: @est31 Is there a more idiomatic way to create an empty `VertexBuffer` (without having to define a `Vertex` type)?

build/main.rs
est31: > My question for you is: What does the version of a registry mean? Does the version of a registry mean that the extensions specified for it will only be generated when the installed OpenGL version is below the version of the registry?

Note that this is in build/main.rs so the code runs at compile time, not runtime. At that point, there is no concept of installed opengl version.

The request for using a specific opengl version is handled in a different part of the code, as is checking for availability of an extension.

Second, I dont know exactly what the minimum opengl version is that glium supports, but it's certainly not 4.5. I don't think  you need two separate registries here, just add the extension and maybe increase the version number in the existing registry. That being said I'm not 100% certain what this number does. gl-generator docs dont show anything, maybe look at the source code of the crate?

examples/screenshot-spirv.rs
est31: Not that I knew. All the examples I've seen do define a vertex type.

Boscop:
> Also would be great to rebase the PR onto latest master, as I can only merge PRs that are in such a state.

Yeah, of course. I just rebased. And I change the code from using a hardcoded entry point ("main") to allowing specifying the entry point name for vertex and fragment shader.
I also added a README.md file for the SPIR-V example, that explains everything.

> Is there a way to make the program compile the spirv files on the fly? Or at least would it be possible to get both spirv files below 1 kb each? I'm not a fan of adding binary files :).

Now it loads both vertex and fragment shader from the same `shader.spv` file that is 1.5 kb (so it's smaller than if both were 1 kb). Is that ok?
Because I think it's best to not require users to have to install `glslangValidator` just to be able to run the SPIR-V example, so the spv file should be included.

> thread 'main' panicked at 'The program attribute `` is missing in the vertex bindings', src/vertex_array_object.rs:287:17

Hm, it seems it's looking for a member of `Vertex` that has an empty name(?)
Maybe it's related to the fact that I use an empty `VertexBuffer` somehow? (But it works for me.)
Could you try again now?
If it still doesn't work, you could try with the same vertex buffer that the `screenshot` example uses.

build/main.rs
Boscop: Ah yeah, I understood it now:
When I only use version `(4, 5)`, then `gl.SpecializeShader` is not generated (because it's part of core OpenGL 4.6), and when I use `(4, 6)` but don't have `"GL_ARB_gl_spirv"` in the extension list, then `gl.SpecializeShaderARB` is not generated.
So I need to have both, version `(4, 6)` and `"GL_ARB_gl_spirv"` in the extension list, then everything works. I updated it in my branch.

est31:
> Hm, it seems it's looking for a member of Vertex that has an empty name(?)
Maybe it's related to the fact that I use an empty VertexBuffer somehow? (But it works for me.)
Could you try again now?
If it still doesn't work, you could try with the same vertex buffer that the screenshot example uses.

I tried all of that. It's the same error in all instances.

est31:
I've tried it on both a Radeon Vega GPU as well as on an Intel HD Graphics 530 (both onboard GPUs). Both using MESA as a driver (20.2.1 in the case of the Intel HW, and 20.1.7 in the case of the Ryzen HW). Both gave that error. What are your drivers?

Boscop:
I'm on Manjaro with a GTX 1060 using the nvidia 4.6 driver. 
Which opengl version do you have with those drivers? And do you have the SPIR-V extension?

(I'm trying to determine if it's an issue with the empty vertex buffer in the example or with the SPIR-V loading code.)

est31:
Just ran the checks on the intel hw: 4.6 is supported, and grepping glxinfo for spirv lists spirv extensions as well. I remember that both devices support vulkan so it'd be weird if spirv didn't work.

willemv:
@est31 Can you try adding following lines in `src/program/reflection.rs`, starting at line 381 (i.e. after the lines that check whether an attribute names starts with gl_) :

```
        if attr_name.is_empty() {   // TODO is this a driver issue we need to work around, or something we're missing?
            continue;
        }```

Are you able to run the example then ?

**Comment by willemv** [comment]
Reading the `glGetActiveAttrib` documentation, it looks like the way glium calls it could be improved.

Specifically:
* `reflect_attributes` in `src/program/reflection.rs` now assumes a max attribute name length of 63 chars. You can, however, request the longest attribute name by calling `glGetProgramiv` with `	GL_ACTIVE_ATTRIBUTE_MAX_LENGTH` first.
* `reflect_attributes` should also deal with error cases. From the docs:
> This function will return as much information as it can about the specified active attribute variable. If no information is available, length will be 0, and name will be an empty string. This situation could occur if this function is called after a link operation that failed. If an error occurs, the return values length, size, type, and name will be unmodified.

This empty string is what you seem to be running into, @est31

**Comment by willemv** [comment]
I've added a pull request that first queries the max attribute name length using `glGetProgramiv` with `GL_ACTIVE_ATTRIBUTE_MAX_LENGTH`

See https://github.com/glium/glium/pull/1898

**Comment by willemv** [comment]
Re. compiling to SPIR-V on the fly : there is Google's shaderc, which has a rustbinding. https://crates.io/crates/shaderc

However, that's a rather large dependency that requires a local installation of CMake and ninja. Adding that dependency to the `spirv` example will make it more complicated for contributors, so I don't think you want to add that dev-dependency. I think the `README.MD` of the `spirv` sample has a good explanation already on how to compile the shaders source to SPIR-V binary, so that could be sufficient. 

Maybe a link to the rust bindings to shaderc might be useful.

**Comment by Boscop** [comment]
@willemv Thanks for looking into this! 

@est31 I've been talking with @msiglreith on discord and he noticed that `glslangValidator` generated this empty attribute:
```llvm
OpEntryPoint Vertex %main "main" %uv %gl_VertexID %_ %gl_InstanceID
OpName %_ ""
```
(Btw, when I used `glslc` instead, it also generates this.)

@msiglreith said:
> looks like the implicit `gl_PerVertex` block..
> somehow it seems to confuse the output block as attribute. I'm also running a nvidia gpu so I don't see this issue as well. Might have more luck with the newer `GetProgramInterface` based api but not sure. In general, name based reflection can become tricky as names are not required for spirv :/
> the empty attribute is the implicit vertex output struct containing the `gl_Position` I believe.

For some reason, it works fine on Nvidia cards but not on others, unless the empty attribute is skipped like @willemv suggests.
> could also try filter out empty string named attributes as you can't match against them anyway as workaround

So yeah, I think the best way to fix this is to just skip the empty attribute.

est31:
> Now it loads both vertex and fragment shader from the same shader.spv file that is 1.5 kb (so it's smaller than if both were 1 kb). Is that ok?

Reconsidering it's not that big after all. I just dont like binary files in general as they bloat up the size of the repo, especially when you re-run some tool regularly to regenerate them with the newest version of that tool... I think in this instance it's fine, as long as you don't plan to make PRs that regenerate them.

Thanks for all the debugging work, you seem to have found the issue. I'll take a closer look at the PR tomorrow.

src/program/mod.rs
```diff
@@ -283,9 +283,108 @@ pub enum ProgramCreationInput<'a> {
...
+    ) -> Self {
+        Self {
+            vertex_shader,
+            fragment_shader,
+            tessellation_control_shader: Default::default(),
+            tessellation_evaluation_shader: Default::default(),
+            geometry_shader: Default::default(),
+            transform_feedback_varyings: Default::default(),
+            outputs_srgb: Default::default(),
+            uses_point_size: Default::default(),
```
est31, line 337: Could you add false, None, etc instead of Default? This would make this code more informative.

est31 (changes requested):
Mostly ok except one nit, and the request to create a clean history, i.e. one where only one spirv file appears. If you don't want to change git history, it's no problem, I'll squash it at merging then.

I've tried out the PR locally and now, as of #1898 merged, it works great.

Regarding the `COMPILER_GLOBAL_LOCK` question, I think it's because some drivers have per-process resources that multiple threads with multiple OpenGL contexts would have issues with. At least that's how I understand commits 9bf4003c23c8469091c35f5fc1b585d82f90d42e and 115277f2f42c9a7e3147af824d6118a69ac48c19 . Since then, glutin's event loop handling has become more strict, and only allows one event loop per process, so maybe such a lock isn't that helpful any more. I think it's better to be on the safe side though and just keep it around.

src/program/mod.rs
Boscop: Sure. Done.

Boscop:
> no problem, I'll squash it at merging then

Thanks!

I pushed the requested changes. Not sure why the CI builds were cancelled. 
Also not sure why macOS fails: If I change the code to how rustc suggests, it doesn't build on my computer (linux).

est31:
> Not sure why the CI builds were cancelled.

If there is a failure in ine of the jobs, Github actions cancels the other jobs that are still running. So it's because of macos.

> If I change the code to how rustc suggests, it doesn't build on my computer (linux).

Hmm that's weird. We'll need something that works portably on all platforms. Which brings up a good point, I think I should add Windows to the CI too.

Have you tried reading how similar code in different places does it? Maybe you need to cast to a different type?

willemv:
Re. macOS build failures: this is probably because `GLhandleARB` is defined differently on macOS vs. other platforms:
```rust
#[cfg(target_os = "macos")]
pub type GLhandleARB = *const super::__gl_imports::raw::c_void;
#[cfg(not(target_os = "macos"))]
pub type GLhandleARB = super::__gl_imports::raw::c_uint;
```
(from the generated `gl_bindings.rs` in the build directory)

The code in `src/program/shaders.rs` looks weird though: you're passing a handle as the `count` parameter. Are you sure that is correct, @Boscop ?

willemv:
It looks like there is no ARB-equivalent for `glShaderBinary` (grepping through the `gl_binding.rs`). Also, the `build_spirv_shader` is written in such a way that `id` is _never_ a handle, but always an ID. I think this means you don't need to check for `Handle::Handle` case in the `match id` branch. I even think you can simplify the method even further and simply make the `id` a `gl::types::GLuint`

For reference:
```rust
        let id = if ctxt.version >= &Version(Api::Gl, 4, 6) || ctxt.extensions.gl_arb_gl_spirv {
            Handle::Id(ctxt.gl.CreateShader(shader_type))
        } else {
            return Err(ProgramCreationError::CompilationNotSupported);
        };

        if id == Handle::Id(0) || id == Handle::Handle(0 as gl::types::GLhandleARB) {
            return Err(ProgramCreationError::ShaderTypeNotSupported);
        }

        let binary = &spirv.binary;
        let entry_point = ffi::CString::new(spirv.entry_point.as_bytes()).unwrap();

        const GL_SHADER_BINARY_FORMAT_SPIR_V_ARB: gl::types::GLenum = 0x9551;
        match id {
            Handle::Id(id) => {
                assert!(ctxt.version >= &Version(Api::Gl, 4, 6));
                ctxt.gl.ShaderBinary(1, &id, GL_SHADER_BINARY_FORMAT_SPIR_V_ARB, binary.as_ptr() as _, binary.len() as _);
            },
            Handle::Handle(id) => {
                assert!(ctxt.extensions.gl_arb_gl_spirv);
                ctxt.gl.ShaderBinary(1, &id, GL_SHADER_BINARY_FORMAT_SPIR_V_ARB, binary.as_ptr() as _, binary.len() as _);
            }
        }
        //...
```

Boscop:
> The code in src/program/shaders.rs looks weird though: you're passing a handle as the count parameter. Are you sure that is correct, @Boscop ?

Where exactly? I'm passing 1 as the count to `ShaderBinary`.

> fn ShaderBinary(&self, count: types::GLsizei, shaders: *const types::GLuint, binaryformat: types::GLenum, binary: *const __gl_imports::raw::c_void, length: types::GLsizei)

willemv:
> > The code in src/program/shaders.rs looks weird though: you're passing a handle as the count parameter. Are you sure that is correct, @Boscop ?
> 
> Where exactly? I'm passing 1 as the count to `ShaderBinary`.

My bad, I miscounted. You're passing the handle as the `shaders` parameter, which is a `GLuint`. So that's not a mistake.

Boscop:
> the build_spirv_shader is written in such a way that id is never a handle, but always an ID.

Maybe that's a mistake though? Maybe I should mirror `build_shader` and write:
```rust
let id = if ctxt.version >= &Version(Api::Gl, 4, 6) {
    Handle::Id(ctxt.gl.CreateShader(shader_type))
} else if ctxt.extensions.gl_arb_gl_spirv {
    Handle::Handle(ctxt.gl.CreateShaderObjectARB(shader_type))
} else {
    return Err(ProgramCreationError::CompilationNotSupported);
};
```
Does `CreateShaderObjectARB` support SPIR-V shaders?

Boscop:
So the problem is that the bindings don't use `#[cfg(target_os = "macos")]` also for generating different versions of `ShaderBinary` and `SpecializeShaderARB`?
Any idea how to make the bindings generator do this?

willemv:
> Does `CreateShaderObjectARB` support SPIR-V shaders?

I don't think the `gl_arb_gl_spirv` extension adds the `CreateShaderObjectARB` method. That method comes from the very old `GL_ARB_shader_objects` extension, which is irrelevant in this context I think.

The `glCreateShader` method is supported since OpenGL 2.0, so if you want to be really precise, I think you precondition can be :|

```
let id = if ctxt.version >= &Version(Api::Gl, 4, 6) || (ctxt.version >= &Version(Api::Gl, 4, 6) && ctxt.extensions.gl_arb_gl_spirv )
```

I.e. if SPIRV is not natively supported, check that the OpenGL version is at least 2.0 _and_ that the `gl_arb_gl_spirv` extension is present.

In any case, the id is _always_ a `GLuint`, so you can remove the code that tries to differentiate between the `Handle::Id` and `Handle::Handle` cases

willemv:
> 
> 
> So the problem is that the bindings don't use `#[cfg(target_os = "macos")]` also for generating different versions of `ShaderBinary` and `SpecializeShaderARB`?
> Any idea how to make the bindings generator do this?

No, I think the binding generator is fine. `glShaderBinary` is a method that exists regardless of the `gl_arb_gl_spirv` extension. You simply should not attempt to call that method when you are in a case where you are using `gl::types::GLhandleARB`

As I tried explaining in my other comments, you're never in this case with the current implementation. Sorry if my comments are not clear enough.

willemv:
From the `GL_ARB_gl_spirv` extension [spec](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_gl_spirv.txt):
> This extension requires version 3.3 or later of The OpenGL Graphics System.
>   (It is not written for OpenGL ES.)

OpenGL 3.3 supports the `GLuint glCreateShader(...)` method. This means there is no need to fall back to a `handleARB glCreateShaderObjectARB(...)` method. That means the `id` variable can never be a `GLhandleARB` and you can simply make it an `GLuint` instead of a `glium::Handle`. That avoids the match statement entirely, so also the `Handle::Handle` branch that now doesn't compile on macOS.

So that addresses the compile error on macOS.

Reading some more about binary shaders, I think this is the situation:

`void glShaderBinary(...)` was introduced in OpenGL 4.1. I can't find an extension that lets people use this in older OpenGL versions, so I think the functionality of this PR simply requires at least OpenGL 4.1.

There is the `ARB_get_program_binary` extension, which adds the `void glProgramBinary(...)` method, but I think that is not quite the same.

So the if should become:
`if ctxt.version >= &Version(Api::Gl, 4, 6) || (ctxt.version >= &Version(Api::Gl, 4, 1) && ctxt.extensions.gl_arb_gl_spirv )`

Boscop:
@willemv Thanks. I pushed my changes. @est31 I guess it can be merged now? :)

est31 (approved):
Thanks for the PR!

PR #1898, opened 2020-12-13
first query the max attr name length before querying attribute names

willemv:
Avoid imposing a max attribute name length of 64 chars. You can query the max attribute name length.

Also, when no information about the attribute is returned, skip it. Note the TODO though: is this really what you want?

As discussed in https://github.com/glium/glium/pull/1895


est31 (approved):
Thanks!

est31:
I've merged this but ideally we'd have a comment explaining that some spirv shaders may create empty attribute names.

willemv:
I just realized I better take `63` as a lower bound for the max attr name length. I.e. let's not trust the driver too much :s

PR #1899, opened 2020-12-14
Clamp max attribute name length to [63, 2047] and comment

willemv:
Let's not trust the driver too much.

### DIFF
--- a/src/program/reflection.rs
+++ b/src/program/reflection.rs
@@ -1,5 +1,6 @@
 use crate::gl;
 
+use std::cmp;
 use std::collections::HashMap;
 use std::hash::BuildHasherDefault;
 use std::ffi;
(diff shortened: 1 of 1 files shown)


est31:
Can you add a comment why you are doing a max? Also, it'd be helpful to add a comment as I described in #1898.

willemv:
Let's also add an upper bound instead of only a lower bound.

willemv:
> 
> 
> Can you add a comment why you are doing a max? Also, it'd be helpful to add a comment as I described in #1898.

will do

src/program/reflection.rs
```diff
@@ -345,6 +346,9 @@ pub unsafe fn reflect_attributes(ctxt: &mut CommandContext<'_>, program: Handle)
                                                 &mut attr_name_len_max);
             }
         };
+        //let's not trust the driver too much, and clamp the max_len to [63, 2047]
```
est31, line 349: Can you add a space between // and the text? This is more commonly used in the codebase.

est31 (changes requested):

src/program/reflection.rs
willemv: Done!

est31 (approved):
Thanks!

PR #1905, opened 2020-12-19
Add missing use to full example

meringu:
This is in an above step but is missing from the full example.


est31:
It's already imported in the main function, no?

meringu:
Sorry, new to rust. I didn't catch that!

est31:
@meringu no problem!

PR #1909, opened 2021-01-26
Make `Content::read()` API `unsafe`

### LINKED ISSUE #1907: Uninitialized buffer exposed to user provided fn
Hello :crab: ,
we (Rust group @sslab-gatech) found a memory-safety/soundness issue in this crate while scanning Rust code on crates.io for potential vulnerabilities.

## Issue Description

https://github.com/glium/glium/blob/3975e7dcf20b238bbc11f3e9a88dd1e4463228e8/src/buffer/mod.rs#L141-L154

`<[T] as buffer::Content>::read` method creates an uninitialized buffer (`value`) and passes it to user-provided fn `F`. 
`F` is allowed to read from the uninitialized buffer.
This is unsound, because it allows safe Rust code to exhibit an undefined behavior (read from uninitialized memory).

[This part](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) from the `Read` trait documentation explains the issue:

> It is your responsibility to make sure that `buf` is initialized before calling `read`. Calling read with an uninitialized `buf` (of the kind one obtains via `MaybeUninit<T>`) is not safe, and can lead to undefined behavior.

Following are also relevant articles regarding the issue
* https://rust-lang.github.io/rfcs/2930-read-buf.html#but-how-bad-are-undefined-values-really
* https://www.ralfj.de/blog/2019/07/14/uninit.html

Thank you for checking out this issue :+1:

JOE1994:
Fixes #1907

Two provided impls of `Content::read()` are safe APIs that can trigger UB in that
they can either (1)expose a zeroed `T` that cannot be `T` or
(2) expose an uninitialized vector of `T` to a user-provided function.
`Content::read()` API can be called from safe Rust to trigger undefined behavior.

I learned from discussion in #1907 that the user-provided function parameter of `Content::read()` is not
expected to read but only expected to write to `&mut Self`,
but I couldn't find this description in the documentation.

This commit switched `Content::read()` API to an `unsafe` API, and also
added warning message to documentation.
Since it is not possible to statically ensure that user-provided function `F`
(parameter of `Content::read()` will not read from `&mut Self`,
making this API an `unsafe` API will impose the responsibility of guaranteeing safety
 to both API implementors and users.

Thank you for reviewing this PR :+1:

### DIFF
--- a/src/buffer/mod.rs
+++ b/src/buffer/mod.rs
@@ -82,7 +82,8 @@ pub unsafe trait Content {
     type Owned;
 
     /// Prepares an output buffer, then turns this buffer into an `Owned`.
-    fn read<F, E>(size: usize, _: F) -> Result<Self::Owned, E>
+    /// User-provided closure `F` must only write to and not read from `&mut Self`.
+    unsafe fn read<F, E>(size: usize, _: F) -> Result<Self::Owned, E>
                   where F: FnOnce(&mut Self) -> Result<(), E>;
 
     /// Returns the size of each element.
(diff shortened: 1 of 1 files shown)


JOE1994:
To refer to an example proof of concept that triggers UB using the `Content::read()` API,
please refer to https://github.com/glium/glium/issues/1907#issuecomment-755469264.

Thank you!

est31 (approved):
Thanks!

est31:
(FTR I want to merge this once a new breaking glutin version is released because it's a breaking change)

Issue #1849, opened 2020-06-04
`UniformsStorage` should be `Copy`/`Clone`


Boscop:
I have a situation where I need to pass the value returned by `uniform!{}` to different sub-renderers, who each add their own uniforms with `let uniforms = uniforms.add("foo", value);`.
These sub-renderers use dynamic uniforms (not known at compile-time, but based on a config loaded at runtime).
If I pass the value returned by `uniform!{}` down by ref, and in the sub-renderer I do `let uniforms = uniforms.add("foo", value);` I get an error that `UniformsStorage` is not `Copy`. If I pass it down by value, I get basically the same error ("use of moved value: `uniforms`. value moved here, in previous iteration of loop") because it's not `Copy` nor `Clone`.
I looked at the code and it seems to be possible to just `derive(Copy, Clone)` so that it's `Copy` whenever `T` and `R` are `Copy`:
```rust
pub struct UniformsStorage<'n, T, R> where T: AsUniformValue, R: Uniforms {
    name: &'n str,
    value: T,
    rest: R,
}
```
Which is basically the case for all types that impl `AsUniformValue` and `Uniforms`!

Could you please `derive(Copy, Clone)` on `UniformsStorage` :)

est31:
PRs welcome!

est31:
#1915 is merged now.

PR #1915, opened 2021-03-03
Made UniformStorage implement Copy/Clone

InnocentusLime:
A pull request for implementing issue #1849.

Decided to make a PR of that, since the feature seems useful and trivial, but the original author of the issue had seemingly disappeared


est31 (approved):
Thanks!

Issue #1888, opened 2020-11-17
Cannot find macro `uniform` in this scope


Kirottu:
So, my project is failing to compile with this error message, regarding my editor the errors origin is `glium-0.28.0/src/macros.rs`

And here is the full compilation error:
```
error: cannot find macro `uniform` in this scope
  --> src/main.rs:72:24
   |
72 |           let uniforms = glium::uniform! {
   |  ________________________^
73 | |             matrix: [
74 | |                 [1.0, 0.0, 0.0, 0.0],
75 | |                 [0.0, 1.0, 0.0, 0.0],
...  |
79 | |             tex: &texture,
80 | |         };
   | |_________^
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error
```

It worked just fine before too, it just spontaneously stopped working.

est31:
Can you try removing code until you get a minimum reproducible example? Alternatively, check whether older versions still compile, maybe you changed something seemingly unrelated and it caused the breakage.

Kirottu:
It seemingly stopped working after i implemented code to load a texture. Which is pretty bizarre, as after that i disabled the texture loading code with nothing changed.

InnocentusLime:
This error persists even on glium 0.29.0. Here's the minimal example

```rust
fn main() {
  glium::uniform! {
    foo : 2,
    bar : 2.0f32,
  };
}
```

This produces the same error.

My guess is that `uniform!` macro doesn't use its absolute path in one of the branches (see [line 72 in `uniform!` definition](https://github.com/glium/glium/blob/master/src/macros.rs#L72))

The fix would be to write 

```rust
$crate::macros::uniform!($($field: $value),*)
```

instead

PR #1916, opened 2021-03-06
Quick fix for #1888

InnocentusLime:
This PR is supposed to fix issue #1888. However, that issue itself was caused by a much bigger problem! The problem is that `glium` hasn't fully moved to Rust 2018 edition's style (at least in case of macros). The doc examples and code are semi-outdate.

@est31 if the fix is urgent, we can merge it right now. As an alternative we can halt this PR and I will write a full migration for macros to Rust 2018!


est31 (approved):
Thanks!

PR #1917, opened 2021-03-06
Fixed other macros

InnocentusLime:
This is a followup to PR #1916. Turns out other macros in `glium` possess the same `vulnerability`, so I added more `$crate::` calls just in case some else uses those macros without importing them


est31 (approved):
Thanks!

PR #1921, opened 2021-04-20
Adopt intra-doc-links

est31:
There were a couple of manual links to documentation. Replace them with intra-doc-links.


est31:
For the build failure, I've filed #1922 .

Issue #1935, opened 2021-06-06
per_instance() incorrectly returns InstancingNotSupported on GLES 3


cosmicchipsocket:
Was busy porting my game engine to Emscripten and noticed that it was erroring out because per_instance() was returning Err(InstancingNotSupported). I looked around and it turns out that WebGL 2 does infact support instancing, so I modified glium to remove the OpenGL version and extension checks and it started working.

The per_instance() functions should return successful if the current context is GLES 3 or newer.

est31:
Fixed by #1936

PR #1936, opened 2021-06-06
Allow instancing on GLES 3 or later

cosmicchipsocket:
This adds additional conditionals to per_instance() to allow instancing on GLES 3 or newer contexts.


cosmicchipsocket:
Fixes #1935

est31 (approved):
Thanks!

PR #1938, opened 2021-06-07
Format files with cargo fmt

fayalalebrun:
I am working on a PR for an unrelated feature, but the fact that the code does not seem to be formatted in a standard way is bothersome.

This just formats all the source files with rust fmt, which is the standard nowadays.


est31:
As of this PR, rustfmt worsens a lot of code, which used to be aligned very well and now isn't. Either one needs to find options or manually turn it off around the affected places.

est31:
Example of the problem:

Screenshot_20210607_142938

New state:

Screenshot_20210607_143009

avl:
I propose we just close this PR. The list of issues and PRs is getting quite long as is.

I agree that rustformat doesn't always make things better, and churn for unclear benefit is the last glium needs. Let's concentrate on the important stuff.

est31:
Yeah the chance is low that rustfmt adopts the beauty standards I want. Closing.

PR #1939, opened 2021-06-08
Add support for importing and using Vulkan semaphores

fayalalebrun:
Thanks to the `EXT_external_objects` OpenGL extension, there now exists a possibility to interoperate between OpenGL and Vulkan. Specifically, this extension allows OpenGL to use Vulkan textures and semaphores.

This PR allows Glium to use Vulkan semaphores imported from linux file descriptors. Next, I will be working on allowing OpenGL to import textures created in Vulkan.

An example showing this working with Vulkano can be found here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=119006cc560b90947054cfb72e80d0ef

The related Vulkano pull request can be found here: https://github.com/vulkano-rs/vulkano/pull/1606

### DIFF
--- a/src/semaphore.rs
+++ b/src/semaphore.rs
@@ -1,0 +1,110 @@
+#![cfg(feature = "vk_interop")]
+
+use std::{fs::File, mem, rc::Rc};
+
+use crate::{Context, ContextExt};
+
+use crate::{backend::Facade, context::CommandContext, gl};
+
+/// Describes an error encountered during semaphore creation
+#[derive(Debug, Clone, Copy)]
+pub enum SemaphoreCreationError {
+    /// Driver does not support EXT_semaphore
+    SemaphoreObjectNotSupported,
+    /// Driver does not support EXT_semaphore_fd
+    SemaphoreObjectFdNotSupported,
+    /// OpenGL returned a null pointer when creating semaphore
+    NullResult,
+}
+
+/// Describes a semaphore which can be used for OpenGL-Vulkan command queue synchronization
+pub struct Semaphore {
+    context: Rc<Context>,
+    id: gl::types::GLuint,
+    backing_fd: Option<File>,
+}
+
+impl Semaphore {
+    /// Creates a semaphore imported from a opaque file descriptor.
+    #[cfg(target_os = "linux")]
+    pub unsafe fn new_from_fd<F: Facade + ?Sized>(
+        facade: &F,
+        fd: File,
+    ) -> Result<Self, SemaphoreCreationError> {
+        use std::os::unix::io::AsRawFd;
+
+        let ctxt = facade.get_context().make_current();
+        let sem = Self::new(facade, &ctxt)?;
+
+        if ctxt.extensions.gl_ext_semaphore_fd {
+            ctxt.gl
+                .ImportSemaphoreFdEXT(sem.id, gl::HANDLE_TYPE_OPAQUE_FD_EXT, fd.as_raw_fd());
+
+            if ctxt.gl.IsSemaphoreEXT(sem.id) == gl::FALSE {
+                Err(SemaphoreCreationError::NullResult)
+            } else {
+                Ok(sem)
+            }
+        } else {
+            Err(SemaphoreCreationError::SemaphoreObjectFdNotSupported)
+        }
+    }
+
+    fn new<F: Facade + ?Sized>(
+        facade: &F,
+        ctxt: &CommandContext<'_>,
+    ) -> Result<Self, SemaphoreCreationError> {
+        if ctxt.extensions.gl_ext_semaphore {
+            let id = unsafe {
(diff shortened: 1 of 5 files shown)


src/semaphore.rs
```diff
@@ -0,0 +1,110 @@
...
+                std::ptr::null(),
+            )
+        }
+    }
+}
+
+impl Drop for Semaphore {
+    fn drop(&mut self) {
+        let ctxt = self.context.get_context().make_current();
+        unsafe { ctxt.gl.DeleteSemaphoresEXT(1, mem::transmute(&self.id)) };
```
est31, line 108: I think casting the pointer via `as` would be better here.

src/semaphore.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /// Sends signal through semaphore.
+    pub fn signal(&self) {
+        let ctxt = self.context.get_context().make_current();
+        unsafe {
+            ctxt.gl.SignalSemaphoreEXT(
+                self.id,
+                0,
+                std::ptr::null(),
```
est31, line 96: What about these data? Why are they hardcoded to null here?

est31 (changes requested):

fayalalebrun:
I added support for the layout transition commands, though I won't be able to test it until I implement the external memory and texture import functionality. This should be fairly feature complete now, the only thing missing now is Windows support, via its specific extensions.

src/semaphore.rs
```diff
@@ -70,41 +121,131 @@ impl Semaphore {
...
+        let (buffer_ids, buffer_num, _) = match buffers {
+            Some(buffs) => {
+                let ids = buffs.iter().map(|b| b.get_id()).collect::<Vec<_>>();
+                (ids.as_ptr(), buffs.len(), Some(ids))
+            }
+            None => (std::ptr::null(), 0, None),
+        };
+
+        let (texture_ids, texture_layouts, textures_num, _, _) = match textures {
+            Some(textures) => {
```
est31, line 206: This should be an `if let`

src/semaphore.rs
```diff
@@ -70,41 +121,131 @@ impl Semaphore {
...
+    /// Before signalling the semaphore, the `textures` are transitioned into the layout specified.
+    pub fn signal<T: ?Sized>(
+        &self,
+        textures: Option<&[(TextureAny, TextureLayout)]>,
+        buffers: Option<&[Buffer<T>]>,
+    ) where
+        T: Content,
+    {
+        let (buffer_ids, buffer_num, _) = match buffers {
+            Some(buffs) => {
```
est31, line 198: This should be an `if let`

src/semaphore.rs
```diff
@@ -70,41 +121,131 @@ impl Semaphore {
...
+        &self,
+        textures: Option<&[(TextureAny, TextureLayout)]>,
+        buffers: Option<&[Buffer<T>]>,
+    ) where
+        T: Content,
+    {
         let ctxt = self.context.get_context().make_current();
+
+        let (buffer_ids, buffer_num, _) = match buffers {
+            Some(buffs) => {
```
est31, line 144: This should be an `if let`

src/semaphore.rs
```diff
@@ -70,41 +121,131 @@ impl Semaphore {
...
+                let ids = buffs.iter().map(|b| b.get_id()).collect::<Vec<_>>();
+                (ids.as_ptr(), buffs.len(), Some(ids))
+            }
+            None => (std::ptr::null(), 0, None),
+        };
+
+        let (texture_ids, texture_layouts, textures_num, _, _) = match textures {
+            Some(textures) => {
```
est31, line 152: This should be an `if let`

est31 (changes requested):

est31:
> This should be fairly feature complete now, the only thing missing now is Windows support, via its specific extensions.

Do you want to add it to this PR or do you want to file a new PR for it? Your choice.

src/semaphore.rs
fayalalebrun: I think in this case the argument could be made that `match` is the better choice, since you are indeed matching exhaustively, and in my opinion in this particular case it is clearer.  Is there some reasoning that I am missing which makes `if let` the superior option here?

Also, have you considered adding clippy to the pipeline in order to catch this type of mistakes? (It doesn't seem to give this recommendation)

fayalalebrun:
> Do you want to add it to this PR or do you want to file a new PR for it? Your choice.

For my use case, Linux support is all I need. Will add a TODO.

src/semaphore.rs
est31: `match` adds two indentation levels while `if let` only adds one.

I'm not a fan of all clippy lints but this suggestion by clippy is good.

est31 (approved):
Thanks!

PR #1940, opened 2021-06-09
Correct formulas in the normal mapping tutorial.

marcianx:
Per http://www.thetenthplanet.de/archives/1180 linked in the tutorial:
- The position should be negated.
- The normal must be normalized.

These corrections are not noticeable in a planar example here, but if someone were to copy-paste this example to apply it to, say, a teapot, and apply a normal mapping without a texture the difference is noticeable:
- Not negating the position makes the edges between the bricks look different than one might have expected.
- The interpolation before the fragment shader can make normals unnormalized.

Also, the demo example was animated to make the effect of normal mapping apparent.

### DIFF
--- a/book/tuto-14-wall.md
+++ b/book/tuto-14-wall.md
@@ -187,7 +187,7 @@ Thanks to this we can calculate the *real* normal, in other words the normal of 
 at the given pixel:
 
 ```glsl
-mat3 tbn = cotangent_frame(v_normal, v_position, v_tex_coords);
+mat3 tbn = cotangent_frame(v_normal, -v_position, v_tex_coords);
 vec3 real_normal = normalize(tbn * -(normal_map * 2.0 - 1.0));
 ```
 
--- a/examples/tutorial-14.rs
+++ b/examples/tutorial-14.rs
@@ -100,8 +100,9 @@ fn main() {
             vec3 diffuse_color = texture(diffuse_tex, v_tex_coords).rgb;
             vec3 ambient_color = diffuse_color * 0.1;
 
+            vec3 v_normal_unit = normalize(v_normal);
             vec3 normal_map = texture(normal_tex, v_tex_coords).rgb;
-            mat3 tbn = cotangent_frame(v_normal, v_position, v_tex_coords);
+            mat3 tbn = cotangent_frame(v_normal_unit, -v_position, v_tex_coords);
             vec3 real_normal = normalize(tbn * -(normal_map * 2.0 - 1.0));
 
             float diffuse = max(dot(real_normal, normalize(u_light)), 0.0);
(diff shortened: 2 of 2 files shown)


marcianx:
Also for a more visual comparison of using the same normal map on a different object where I improvised the (u,v) via projection (the CW and CCW orientation of uv doesn't make any difference in the qualitative behavior shown here):

- `v_position` (before)
image
- `-v_position` (after)
image

est31:
@marcianx oh nice the screenshot looks way better than the current tutorial 14. Maybe it can be replaced by the teapot?

est31:
Hmm or actually changing the wall to a teapot would require some major refactoring of the tutorial. Maybe only slightly rotate the wall?

marcianx:
One of the reasons the non-animated teapot makes the normal map more obvious is that there's no texture to distract from the normal mapping. This is what the rectangle looks like in the same color:
image image
However, now that it's animated, the 3d effect of the normal map looks very obvious.

I'm up for more improvements, or even a teapot reuse refactor, but since the teapot doesn't have a native uv map (I used `[(x + 50.) / 100., (y + 50.) / 100.]` with `SamplerWrapFunction::Repeat`), the rectangle seems to be a more easy-to-follow example for the purposes of a tutorial. What do you think?

est31:
@marcianx good point aobut the teapot making it way more complicated. Let's go with the rectangle.

est31:
Sorry about leaving this open for so long, the main reason I did so because the tutorial does not have a healthy fps for me (multiple seconds between frames), so I can't check whether the turning logic is well calibrated. This bug prevents me from trying your code. Does it render nicely for you and update in a good speed?

marcianx:
Really? I wonder why you're seeing such low frame rate. It's running very smoothly for me compiled even in _debug_ mode. This is what I see:
normal_mapping

est31:
Yeah it's weird to me too. Anyways, I think something that's independent from the FPS one ends up having would be better. 

E.g.:

```diff
@@ -118,7 +118,7 @@ fn main() {
     let program = glium::Program::from_source(&display, vertex_shader_src, fragment_shader_src,
                                               None).unwrap();
 
-    let mut t = 0.0f32;
+    let start = std::time::Instant::now();
     event_loop.run(move |event, _, control_flow| {
         let next_frame_time = std::time::Instant::now() +
             std::time::Duration::from_nanos(16_666_667);
@@ -143,7 +143,7 @@ fn main() {
         let mut target = display.draw();
         target.clear_color_and_depth((0.0, 0.0, 1.0, 1.0), 1.0);
 
-        t += 0.03;
+        let t = (std::time::Instant::now() - start).as_secs_f32();
         let ang = t.sin();
         let (c, s) = (ang.cos(), ang.sin());
         let model = [
```

Can you change it accordingly? Then I'll just trust you and merge it.

marcianx:
Sure. I'll need a couple days before I can get back to this. But will do.

est31:
@marcianx take your time!

marcianx:
Thanks! I updated it per your suggestion, but made it animate twice as fast. Feel free to merge. It requires your approval to run the pre-submit workflows.

est31 (approved):
Thanks!

PR #1943, opened 2021-06-23
Add `ClientFormat::U1U5U5U5Reversed`

Amjad50:
This format maps to `UNSIGNED_SHORT_1_5_5_5_REV` glenum.

The library does not support reading `u1u5u5u5` colors values from
`u16`. Adding this formats will add support for this particular input
format.

This color format is used in the PSX GPU, so this support will ease
emulation of such GPU.

```txt
16bit Texture
  0-4   Red       (0..31)
  5-9   Green     (0..31)
  10-14 Blue      (0..31)
  15    Semi Transparency Flag
```


est31 (approved):
Thanks!

PR #1946, opened 2021-07-10
Update ouroboros [fix potential UB]

joshua-maros:
A potential source of undefined behavior was discovered in `ouroboros`. This PR updates it to the latest version `0.10.0` which fixes the problem. The update also drops the requirement that referenced fields exist in a `Box` or another container, so this PR also unboxes those fields.


est31 (approved):
Thanks!

Issue #1775, opened 2019-08-16
invalid_value lint firing


est31:
It seems that @RalfJung's new `invalid_value` lint has found some UB in this crate. I extracted this from the travis CI log of a recent PR:

```
warning: the type `&T` does not permit zero-initialization

   --> src/buffer/view.rs:619:24

    |

619 |         let data: &T = mem::zeroed();

    |                        ^^^^^^^^^^^^^

    |                        |

    |                        this code causes undefined behavior when executed

    |                        help: use `MaybeUninit<T>` instead

    |

    = note: `#[warn(invalid_value)]` on by default

    = note: References must be non-null

warning: the type `&T` does not permit zero-initialization

   --> src/buffer/view.rs:976:24

    |

976 |         let data: &T = mem::zeroed();

    |                        ^^^^^^^^^^^^^

    |                        |

    |                        this code causes undefined behavior when executed

    |                        help: use `MaybeUninit<T>` instead

    |

    = note: References must be non-null

warning: the type `&index::multidraw::DrawCommandNoIndices` does not permit zero-initialization

   --> src/macros.rs:404:57

    |

404 |                       let dummy: &$struct_name = unsafe { mem::zeroed() };

    |                                                           ^^^^^^^^^^^^^

    |                                                           |

    |                                                           this code causes undefined behavior when executed

    |                                                           help: use `MaybeUninit<T>` instead

    | 

   ::: src/index/multidraw.rs:26:1

    |

26  | / implement_uniform_block!(DrawCommandNoIndices, count, instance_count,

27  | |                          first_index, base_instance);

    ;5;12m| |_____________________________________________________- in this macro invocation

    |

    = note: References must be non-null

warning: the type `&index::multidraw::DrawCommandIndices` does not permit zero-initialization

   --> src/macros.rs:404:57

    |

404 |                       let dummy: &$struct_name = unsafe { mem::zeroed() };

    |                                                           ^^^^^^^^^^^^^

    |                                                           |

    |                                                           this code causes undefined behavior when executed

    |                                                           help: use `MaybeUninit<T>` instead

    | 

   ::: src/index/multidraw.rs:45:1

    |

45  | / implement_uniform_block!(DrawCommandIndices, count, instance_count, first_index,

46  | | m                         base_vertex, base_instance);

    | |_____________________________________________________- in this macro invocation

    |

    = note: References must be non-null

warning: the type `&program::compute::ComputeCommand` does not permit zero-initialization

   --> src/macros.rs:404:57

[...]

PR #1950, opened 2021-07-17
Fix implement_buffer_content

avl:
Fix implement_buffer_content so it works on rustc after 2021-03-04. This rustc started using LLVM 12, which miscompiled the 0-references which were created. Code is still UB, but works (for now).

Also, fix the macro so it correctly marks the read fn as unsafe. Without this fix, code using the macro won't compile.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -204,7 +204,7 @@ macro_rules! implement_buffer_content {
                 type Owned = Box<$struct_name<$($gs)*>>;
 
                 #[inline]
-                fn read<F, E>(size: usize, f: F) -> ::std::result::Result<Box<$struct_name<$($gs)*>>, E>
+                unsafe fn read<F, E>(size: usize, f: F) -> ::std::result::Result<Box<$struct_name<$($gs)*>>, E>
                               where F: FnOnce(&mut $struct_name<$($gs)*>) -> ::std::result::Result<(), E>
                 {
                     use std::mem;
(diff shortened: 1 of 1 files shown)


est31:
Thanks but I don't feel this is a fix for the bug. Yes, LLVM might not cause issues in this instance any more, but the pointer to the made up address is still invalid. I wonder if there is a way to do this without creating UB.

avl:
> Thanks but I don't feel this is a fix for the bug. Yes, LLVM might not cause issues in this instance any more, but the pointer to the made up address is still invalid. I wonder if there is a way to do this without creating UB.

I totally agree. This fix isn't especially satisfying. To me it's better than nothing, since it means my product works at all.

I don't know of an easy 'real' fix. One way might be to redesign the API so that the user must provide 'example-instances' of the buffer object, one with 0 elements and one with 1 element. It might be possible to do something with proc-macros, generating code which initializes 'sample' instances which can then be analyzed to figure out sizes.

Another way with proc-macros might be to attempt to recreate the rust layout of the struct, and then calculate the sizes ourselves. Perhaps requiring repr(C). But it seems fragile.

Perhaps the macro should just be removed? I feel it's not good to keep it around when using it can lead to random segfaults for the user.

est31:
Yeah the `implement_uniform_block` macro is likely broken too (see #1775)... maybe we should remove the macros until we can come up with a fix?

coolreader18:
I think a fully correct implementation would involve `std::ptr::metadata` + `std::mem::size_of_val_raw`, or maybe a derive macro. I can't think of any other way to ensure that the struct is an unsized wrapper around a slice (turning 0 into the vtable ptr of a dyn Trait is UB too, since it's a pointer)

avl:
Maybe we could use std::ptr::from_raw_parts somehow? It can be used to create a pointer which I think is not UB. If only we could then somehow get the size of the pointed-to value, we'd have a solution. 

Another idea: Maybe we could modify the macro so rhe user needs to provide the name and type of the last field of the struct? Then we could just use the offset of the last field and the size_of its type?

CodesInChaos:
`std::ptr::from_raw_parts` is unstable. So you either need to require nightly, or rely on unstable implementation details and write similar code yourself.

`from_raw_parts` effectively transmutes the following struct to a fat pointer:
```
#[repr(C)]
struct PtrComponents {
    pub data_address: *const (),
    pub metadata: usize, // since this DST is slice based
}
```

est31:
As my earlier comment writes, changing the address isn't really fixing the issue. Closing in favour of #1951.

Issue #1949, opened 2021-07-17
Undefined behavior in implement_buffer_content!


avl:
The macro implement_buffer_content! contains the following code:

```rust
#[inline]
fn is_size_suitable(size: usize) -> bool {
    use std::mem;

    let fake_ptr: &$struct_name = unsafe { mem::transmute((0usize, 0usize)) };
    let min_size = mem::size_of_val(fake_ptr);

    let fake_ptr: &$struct_name = unsafe { mem::transmute((0usize, 1usize)) };
    let step = mem::size_of_val(fake_ptr) - min_size;

    size > min_size && (size - min_size) % step == 0
}

```

This is, as I understand, undefined behavior. It sort of works by accident in some/many cases. But with recent (>2021-03-04) rustc it does not work in the app I'm developing.

I've tried to solicit help from reddit: https://www.reddit.com/r/rust/comments/om6xs6/help_needed_fixing_ub_in_glium_graphics_library/

I've also made a github repo with a minimal(ish) test case: https://github.com/avl/rust-llvm12-regression

I'm trying to come up with a PR to fix this. I'm very interested in if others have the same problem.

avl:
Here is a PR: https://github.com/glium/glium/pull/1950
With this PR the code works, but the implement_buffer_content-macro is still UB. Just not as much UB :-) .

est31:
Alternative PR: #1951

PR #1951, opened 2021-07-18
Fix miscompile of implement_buffer_content-macro, other approach

avl:
This is an alternative approach to solving the problem https://github.com/glium/glium/issues/1949 .

I think it's 100% free of UB and totally sound. But it only works on nightly rust.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -180,6 +180,17 @@ macro_rules! implement_vertex {
 ///
 /// **Only use this macro on structs.** Using it with anything else will result in a segfault.
 ///
+/// Unfortunately this macro only works on nightly rust. Previously, it was available even on
+/// stable, but the way it was implemented caused undefined behavior and possible miscompiles
+/// on more recent rustc-versions.
+///
+/// To use this macro, you  must place the following at your crate root:
+///
+/// ```
+/// #![feature(layout_for_ptr)]
+/// #![feature(ptr_metadata)]
+/// ```
+///
 /// ## Example
 ///
 /// ```
@@ -204,7 +215,7 @@ macro_rules! implement_buffer_content {
                 type Owned = Box<$struct_name<$($gs)*>>;
 
                 #[inline]
-                fn read<F, E>(size: usize, f: F) -> ::std::result::Result<Box<$struct_name<$($gs)*>>, E>
+                unsafe fn read<F, E>(size: usize, f: F) -> ::std::result::Result<Box<$struct_name<$($gs)*>>, E>
                               where F: FnOnce(&mut $struct_name<$($gs)*>) -> ::std::result::Result<(), E>
                 {
                     use std::mem;
@@ -224,8 +235,9 @@ macro_rules! implement_buffer_content {
                 fn get_elements_size() -> usize {
                     use std::mem;
 
-                    let fake_ptr: &$struct_name = unsafe { mem::transmute((0usize, 0usize)) };
-                    mem::size_of_val(fake_ptr)
+                    let fake_ptr: *const $struct_name = std::ptr::from_raw_parts::<$struct_name>(std::ptr::null(), 0);
+                    // Safety: The metadata-part must be an initialized integer. It is a literal 0.
+                    unsafe { mem::size_of_val_raw(fake_ptr) }
                 }
 
                 #[inline]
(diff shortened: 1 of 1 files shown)


est31:
I like this one much better than the other PR. Not sure if it fixes the UB though. cc @ralfjung is this code UB free now?

RalfJung:
I'm afraid I am lacking context, and it is extremely hard to say what soundness for a macro even means.

Is there some self-contained piece of code (e.g. on the playground) that is representative for what happens here?

src/macros.rs
```diff
@@ -224,8 +235,9 @@ macro_rules! implement_buffer_content {
                 fn get_elements_size() -> usize {
                     use std::mem;
 
-                    let fake_ptr: &$struct_name = unsafe { mem::transmute((0usize, 0usize)) };
-                    mem::size_of_val(fake_ptr)
+                    let fake_ptr: *const $struct_name = std::ptr::from_raw_parts::<$struct_name>(std::ptr::null(), 0);
```
RalfJung, line 238: This is suspicious. The vtable must be non-null even in a raw pointer. In fact it [must even be a valid pointer](https://doc.rust-lang.org/reference/behavior-considered-undefined.html):

> The following values are invalid (at their respective type): [...]
> Invalid metadata in a wide reference, Box<T>, or raw pointer:
> dyn Trait metadata is invalid if it is not a pointer to a vtable for Trait that matches the actual dynamic trait the pointer or reference points to.

RalfJung: I think `std::ptr::from_raw_parts` might be unsound in its current form.

RalfJung: Unless maybe these are all slices? But then you can use the stable `ptr::slice_from_raw_parts`

RalfJung: Okay I guess these must be all slices; for trait objects the metadata has type `DynMetadata` and that cannot be 0.

This is exactly why I need context -- it is very hard to understand macro code from a small patch like this, and I can't spend hours trying to figure this all out myself.^^

est31: Yeah it's a macro to implement custom buffer types. Those buffer types can take up any content. Copy-pasting the rustdoc:
> ## Unsized types
> 
> In order to put some data in a buffer, it must implement the `Content` trait. This trait is
> automatically implemented on all `Sized` types and on slices (like `[u8]`). This means that
> you can create a `Buffer<Foo>` (if `Foo` is sized) or a `Buffer<[u8]>` for example without
> worrying about it.
> 
> However unsized structs don't automatically implement this trait and you must call the
> `implement_buffer_content!` macro on them. You must then use the `empty_unsized` constructor.
> 
> ```no_run
> # #[macro_use] extern crate glium;
> # fn example(display: glium::Display) {
> # use std::mem;
> # use glium::buffer::{BufferType, BufferMode};
> struct Data {
>     data: [f32],        // `[f32]` is unsized, therefore `Data` is unsized too
> }
> 
> implement_buffer_content!(Data);    // without this, you can't put `Data` in a glium buffer

I'm not sure what the use case of this macro is. The documentation contains nothing more than custom wrappers of unsized slices. @avl what are your use cases for it? Maybe the best option is to remove it.

avl: My use case is a buffer which is a DST with some "header info" and then a list of elements. The number of elements is only known at runtime. 

I think I could just as well use two different buffer-objects instead. I would have if I had known that this feature was problematic 😀. But the last couple of years the code has just worked.

avl: > Okay I guess these must be all slices; for trait objects the metadata has type `DynMetadata` and that cannot be 0.
> 
> This is exactly why I need context -- it is very hard to understand macro code from a small patch like this, and I can't spend hours trying to figure this all out myself.^^

Here's a small self-contained example. I've read the rust docs and I think it should be sound. Unless there's some undocumented requirement somewhere.

```rust
#![feature(ptr_metadata)]
#![feature(layout_for_ptr)]

struct Packet {
    sequence_number: u32,     //total 16 byte header
    source_addres: u32,       
    destination_address: u32,
    size: u32,
    data_chunks: [u64]        // arbitrary number of 8 byte payload chunks
}

fn main() {
    let fake_ptr: *const Packet = std::ptr::from_raw_parts::<Packet>(std::ptr::null(), 0);
    // Safety: The metadata-part must be an initialized integer. It is a literal 0.
    let min_size = unsafe { std::mem::size_of_val_raw(fake_ptr) };

    let fake_ptr: *const Packet = std::ptr::from_raw_parts::<Packet>(std::ptr::null(), 1);
    // Safety: The metadata-part must be an initialized integer. It is a literal 1.
    let step = unsafe { std::mem::size_of_val_raw(fake_ptr) - min_size };
    
    // prints: min_size: 16, step: 8
    println!("min_size: {}, step: {}", min_size, step);   
}

```

RalfJung: I'm not very familiar with the new `ptr_metadata` API.

But if you only use this for slices / types where the "unsized tail" is a slice, then I agree you are fine here. This is basically just generalizing `slice_from_raw_parts` from slices to arbitrary types with a slice as their tail.

est31:
As it seems there are valid use cases for this macro and it seems that this PR is not unsound, I'd like to merge this PR. However, it's using the [pointer metadata APIs](https://github.com/rust-lang/rust/issues/81513) which are unstable. Thus the PR can't be merged at the moment.

I think due to the miscompilation I'll remove the macro for the time being and if the features stabilize, this PR can be merged (or feel free to close it and refile later).

Issue #1850, opened 2020-06-04
Lifetime for `UniformValue` of `Sampler` is too strict


Boscop:
Currently `AsUniformValue for Sampler<'a, Texture2d>` is impld [like this](https://docs.rs/glium/0.27.0/src/glium/opt/rustwide/target/debug/build/glium-0681d5e0c6368732/out/textures.rs.html#2807-2809):
```rust
impl<'a> AsUniformValue for Sampler<'a, Texture2d> {
    #[inline]
    fn as_uniform_value(&self) -> UniformValue {
        UniformValue::Texture2d(self.0, Some(self.1))
    }
}
// https://docs.rs/glium/0.27.0/glium/uniforms/enum.UniformValue.html#variant.Texture2d
```

The inferred lifetime for `UniformValue` is that of `self`.
But that is too narrow, instead, it should use the lifetime of `self.0`!
`fn as_uniform_value(&self) -> UniformValue<'a> {`.

In my real-world code this is currently causing
> error: temporary value dropped while borrowed
> creates a temporary which is freed while still in use

But only the `Sampler` is dropped, NOT the `Texture2d`.

est31:
PR welcome!

PR #1956, opened 2021-08-10
Also implement AsUniformValue for textures directly

### LINKED ISSUE #1054: Make AsUniformValue::as_uniform_value() take self by value instead of reference
Taking self by reference makes this API unergonomic due to lifetime issues.

### LINKED ISSUE #1400: Rc<Texture> can
I try implement own uniform type like this:
```rust
extern crate glium;

use glium::texture::Texture2d;
use glium::uniforms::{AsUniformValue, Uniforms};
use std::rc::Rc;


pub struct Foobar {
    pub texture: Rc<Texture2d>,
}

impl Uniforms for DefaultUniforms {
    fn visit_values<'a, F>(&'a self, mut f: F)
        where F: FnMut(&str, UniformValue<'a>)
    {
        f("tex", self.texture.as_ref().as_uniform_value());
    }
}
```
Encountered a lifetime issue, because creat texuture uniform value
```rust
    impl<'a> AsUniformValue for &'a Texture2d
```
Need `&'a Texture`, texture in `Rc` can't meet the lifetime requirements.

but

```rust
f("tex", UniformValue::Texture2d(&*texture, None));
```

Is compile success, because this is `&'a Texture` **into** `UniformValue<'a>`.

strohel:
In addition to implementing if for references, i.e.
`impl<'a> AsUniformValue for &'a {TextureType}`.

The method `as_uniform_value(&self)` takes self by reference, so there
should be no need to require the second-level reference.

This allows a longer lifetime to be automatically inferred. Fixes
compilation of the texture_2d_as_uniform_value_lifetime() test from the
previous commit.

This commit doesn't remove the original impl above in order not to introduce
breaking changes. Removing would mean every texture passing in `uniform! {}`
would have to change.

Should fix #1400. We should also be able to close #1054 (resolved in a slightly different manner).
Related closed issues/PRs: #1629, #1632.

Also related: #1850 - not fixed in this PR, but a similar change to impl for `Sampler` could fix it.
CC @slightknack.

### DIFF
--- a/build/textures.rs
+++ b/build/textures.rs
@@ -425,6 +425,13 @@ fn build_texture<W: Write>(dest: &mut W, ty: TextureType, dimensions: TextureDim
                                 }}
                             }}
 
+                            impl AsUniformValue for {myname} {{
+                                #[inline]
+                                fn as_uniform_value(&self) -> UniformValue {{
+                                    UniformValue::{myname}(self, None)
+                                }}
+                            }}
+
                             impl<'a> AsUniformValue for Sampler<'a, {myname}> {{
                                 #[inline]
                                 fn as_uniform_value(&self) -> UniformValue {{
--- a/tests/texture_creation.rs
+++ b/tests/texture_creation.rs
@@ -87,6 +87,29 @@ fn texture_2d_creation() {
 }
 
 #[test]
+fn texture_2d_as_uniform_value_lifetime() {
+    let display = support::build_display();
+
+    let texture = glium::texture::Texture2d::new(&display, vec![
+        vec![(0, 0, 0, 0), (0, 0, 0, 0)],
+        vec![(0, 0, 0, 0), (0, 0, 0, 0)],
+        vec![(0, 0, 0, 0), (0, 0, 0, 0u8)],
+    ]).unwrap();
+
+    // A function that takes texture reference should be able to return UniformValue (with lifetime
+    // inherited from the reference).
+    fn get_uniforms(texture: &glium::texture::Texture2d) -> glium::uniforms::UniformValue {
+        use glium::uniforms::AsUniformValue;
+        texture.as_uniform_value()
+    }
+
+    let uniforms = get_uniforms(&texture);
+    assert!(matches!(uniforms, glium::uniforms::UniformValue::Texture2d(..)));
+
+    display.assert_no_error(None);
+}
+
+#[test]
 fn empty_texture2d_u8u8u8u8() {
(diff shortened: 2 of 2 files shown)


slightknack:
Thank you @strohel! We're currently using this fork for a portion our pipeline, this change makes it easier to work with uniforms in an idiomatic Rust manner. Let's get this merged!

est31:
> This commit doesn't remove the original impl above in order not to introduce
breaking changes. Removing would mean every texture passing in uniform! {}
would have to change.

Why would it need a change? As far as I'm aware, when a Trait is implemented for T, then it's also implemented for &'a T, no?

strohel:
> Why would it need a change? As far as I'm aware, when a Trait is implemented for T, then it's also implemented for &'a T, no?

Yes when used with a dot operator (method call resolution), but not when used as other function argument, per https://doc.rust-lang.org/nightly/nomicon/coercions.html:
> Note that we do not perform coercions when matching traits (except for receivers, see the next page). If there is an impl for some type U and T coerces to U, that does not constitute an implementation for T. For example, the following will not type check, even though it is OK to coerce t to &T and there is an impl for &T:
> (...)

The culprit here is `impl<'n, T> UniformsStorage<'n, T, EmptyUniforms> where T: AsUniformValue` { `pub fn new(name: &'n str, value: T)` } used from `uniform!{}`. If I remove the `AsUniformValue` impl for references, I get:
```
error[E0277]: the trait bound `&glium::texture::CompressedSrgbTexture2d: AsUniformValue` is not satisfied
   --> examples/image.rs:148:18
    |
148 |             tex: &opengl_texture
    |                  -^^^^^^^^^^^^^^
    |                  |
    |                  the trait `AsUniformValue` is not implemented for `&glium::texture::CompressedSrgbTexture2d`
    |                  help: consider removing the leading `&`-reference
    |
    = help: the following implementations were found:
              <glium::texture::CompressedSrgbTexture2d as AsUniformValue>
```
Passing textures by value might not be an option when the caller does not want to lose ownership.

---
If anyone has ideas how to resolve this without keeping the 2 similar AsUniformValue implementations, I'll happily try them out.

est31:
@strohel fair enough.

est31 (approved):
Thanks!

strohel:
Thanks, @est31, that was quick!

strohel:
@est31, can we please request cutting a patch release with this fix? It would allow us to publish a package that depends on it.

est31:
@strohel sure, will do!

PR #1961, opened 2021-09-12
Fix texture tutorial color space

dcz-self:
The texure was originally loaded in linear space. Substituting the original image for a photo makes it clear that the colors are off, so using SRGB.


est31 (approved):
Great catch!

PR #1965, opened 2021-10-23
Add support for importing Vulkan textures

fayalalebrun:
As a continuation of my previous work, where I added semaphore import  in #1939, I've been working on adding the possibility to import Vulkan textures into Glium.

This MR adds the functionality of the [EXT_external_objects](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_external_objects.txt) into Glium.

In Vulkan, textures and memory are separate concepts. Thus, in order to conform to this, the aforementioned extension introduces the concept of memory objects in OpenGL. These external memory objects can be used to create OpenGL textures. Thus, I've added the `new_from_fd` function to textures, which imports a memory object from an opaque file descriptor and uses that to create GL texture.

With this functionality, it is now possible to import Vulkan textures, and with the existing Vulkan semaphore support, it is possible to synchronize access to these textures.

The only remaining issue at the moment is a crash happening with the Intel Iris driver, when flushing the signal semaphore command from the OpenGL side. Running a debug version of the driver, it prints out `Failed to submit batchbuffer: Invalid argument`. Given that the driver does not pass the piglit `vk-semaphores-2` test, this is probably a problem within the Iris driver. I will submit a bug report to the Mesa repository.

I have created some [sample code](https://gist.github.com/fayalalebrun/8845ba4464af687895cf49ba96b105d6). This program draws to a shared texture with Glium from another thread, which is the drawn to the window by Vulkano. This texture oscillates between a black and green color. I'm interested to hear how well it works on Nvidia or with other Intel/AMD drivers.

[Related Vulkano MR](https://github.com/vulkano-rs/vulkano/pull/1734)

### DIFF
--- a/src/memory_object.rs
+++ b/src/memory_object.rs
@@ -30,12 +30,12 @@ impl std::fmt::Display for MemoryObjectCreationError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         use self::MemoryObjectCreationError::*;
 
-	let desc = match *self {
-	    MemoryObjectNotSupported => "Driver does not support EXT_memory_object",
-	    MemoryObjectFdNotSupported => "Driver does not support EXT_memory_object_fd",
-	    NullResult => "OpenGL returned a null pointer when creating memory object",
-	};
-	f.write_str(desc)
+        let desc = match *self {
+            MemoryObjectNotSupported => "Driver does not support EXT_memory_object",
+            MemoryObjectFdNotSupported => "Driver does not support EXT_memory_object_fd",
+            NullResult => "OpenGL returned a null pointer when creating memory object",
+        };
+        f.write_str(desc)
     }
 }
 
@@ -60,7 +60,7 @@ impl MemoryObject {
         fd: std::fs::File,
         size: u64,
     ) -> Result<Self, MemoryObjectCreationError> {
-	use std::os::unix::io::AsRawFd;
+        use std::os::unix::io::AsRawFd;
         let ctxt = facade.get_context().make_current();
         let mem_obj: Self = Self::new(facade, &ctxt)?;
 
--- a/src/memory_object.rs
+++ b/src/memory_object.rs
@@ -13,6 +13,7 @@ use crate::ContextExt;
 use std::{fs::File, mem, rc::Rc};
 
 /// Describes an error encountered during memory object creation
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum MemoryObjectCreationError {
     /// Driver does not support EXT_memory_object
     MemoryObjectNotSupported,
@@ -21,6 +22,21 @@ pub enum MemoryObjectCreationError {
     /// OpenGL returned a null pointer when creating memory object
     NullResult,
 }
+
+impl std::fmt::Display for MemoryObjectCreationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use self::MemoryObjectCreationError::*;
+
+	let desc = match *self {
+	    MemoryObjectNotSupported => "Driver does not support EXT_memory_object",
+	    MemoryObjectFdNotSupported => "Driver does not support EXT_memory_object_fd",
+	    NullResult => "OpenGL returned a null pointer when creating memory object",
+	};
+	f.write_str(desc)
+    }
+}
+
+impl std::error::Error for MemoryObjectCreationError {}
 
 /// Describes a memory object created by an external API. In OpenGL there is no distinction
 /// between a texture or buffer and its underlying memory. However, in other API's like Vulkan
--- a/src/memory_object.rs
+++ b/src/memory_object.rs
@@ -13,7 +13,7 @@ use crate::version::Version;
 use crate::backend::Facade;
 use crate::context::Context;
 use crate::ContextExt;
-use std::{fs::File, rc::Rc};
+use std::rc::Rc;
 
 /// Describes an error encountered during memory object creation
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


src/texture/any.rs
```diff
@@ -451,7 +456,8 @@ pub fn new_texture<'a, F: ?Sized, P>(facade: &F, format: TextureFormatRequest,
         ty,
         levels: texture_levels as u32,
         generate_mipmaps: should_generate_mipmaps,
-        owned: true
+        owned: true,
+	memory: None
```
est31, line 460: What happened here?

est31:
Thanks for filing the PR and before giving it a close look, it seems that the indentation is quite messed up. Would it be possible to fix that?

fayalalebrun:
@est31 I'm not sure what's going on here. If you look at the diff on github, the indentation looks messed up. However, if you choose "View file", or also on my local editor, the indentation looks fine. Have you encountered this issue before? Perhaps it's related to the use of tabs or spaces?

On a related note, is there any reason why the textures are not generated via the use of macros? Would such a change be a welcome contribution? Doing that would open the door to the use of rust-fmt, avoiding issues like this one.

fayalalebrun:
I've fixed it. The problem was that I had tabs in some places, and the github diff doesn't display that well. However my question about generating the texture source files via macros still remains.

src/memory_object.rs
```diff
@@ -0,0 +1,137 @@
...
+            || ctxt.version >= &Version(Api::GlEs, 3, 2)
+            || ctxt.extensions.gl_arb_texture_storage)
+            && ctxt.extensions.gl_ext_memory_object
+        {
+            let id = unsafe {
+                let mut id: gl::types::GLuint = 0;
+                ctxt.gl.CreateMemoryObjectsEXT(1, &mut id as *mut u32);
+
+                if ctxt.gl.IsMemoryObjectEXT(id) == gl::FALSE {
+                    Err(MemoryObjectCreationError::NullResult)
```
est31, line 107: You can just do `return Err` here, no need for Err and then the ? operator.

src/texture/any.rs
```diff
@@ -481,10 +487,136 @@ pub unsafe fn from_id<F: Facade + ?Sized>(facade: &F,
...
+        }
+
+        if should_generate_mipmaps {
+            generate_mipmaps(&ctxt, bind_point);
+        }
+
+        id
+    };
+    
+    Ok (TextureAny {
```
est31, line 607: This has an extra space after the `Ok` that is not needed.

est31 (changes requested):

est31:
> On a related note, is there any reason why the textures are not generated via the use of macros? Would such a change be a welcome contribution?

The texture generation would ideally use macros yes, but it feels to me that it's easier to do it myself than to review PRs for it, so I'd like to do it myself I think.

src/semaphore.rs
```diff
@@ -72,16 +86,15 @@ impl Into<crate::gl::types::GLenum> for TextureLayout {
...
+    id: gl::types::GLuint,    
 }
 
 impl Semaphore {
     /// Creates a semaphore imported from an opaque file descriptor.
     #[cfg(target_os = "linux")]
     pub unsafe fn new_from_fd<F: Facade + ?Sized>(
         facade: &F,
-        fd: File,
+        fd: std::fs::File,
```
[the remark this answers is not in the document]
fayalalebrun: The problem is that since this function is specific to Linux (Windows requires another extension to be used), if I import this for the entire module the Windows and MacOS builds will have warnings. In order to avoid that I either have to import this only for the function or import it for the entire module and add an annotation to ignore the warning. Which solution would be preferable?

fayalalebrun:
> The texture generation would ideally use macros yes, but it feels to me that it's easier to do it myself than to review PRs for it, so I'd like to do it myself I think.

Good to hear that. It's exactly what macros are meant to help with, and I think if there was some kind of format checking it would be a great time saver. For example, in this case three out of five requested changes had to do with formatting. These could easily have been automated with a pipeline check.

src/semaphore.rs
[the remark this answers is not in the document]
est31: That's a good point. Resolved.

est31 (approved):
Thanks!

est31:
I've checked the adoption of macros. It seems it's going to need the ability concatenate identifiers.

fayalalebrun:
> I've checked the adoption of macros. It seems it's going to need the ability concatenate identifiers.

Have you looked at the [paste](https://github.com/dtolnay/paste.git) crate? It's very popular, and does exactly that.

est31:
I don't like adding a new dependency just for making code a little bit more neat. If paste were in glium's list of (transitive) dependencies already, it would have been different, but it isn't.

fayalalebrun:
I think it's a good practice to keep dependencies to a minimum. However, in this case the crate doesn't pull in any other dependencies, so it's quite light in that regard. Also, the identifier concatenation is impossible to do in a declarative macro, so unless you're planning to do everything a procedural macro, you would have to revive the "macros" sub-crate and basically re-implement that crate's functionality there.

That would be my argument towards considering the use of that crate. However, I think whatever is simplest and easiest to maintain is always best.

yshui:
@fayalalebrun From `EXT_external_object_fd`:

```
        A successful import operation transfers ownership
        of <fd> to the GL implementation, and performing any operation on
        <fd> in the application after an import results in undefined
        behavior.
```

But `new_from_fd` use `as_raw_fd`, not `into_raw_fd`, meaning the file descriptor is going to be closed when the `File` is dropped. That feels wrong?

yshui:
@fayalalebrun Thanks for the clarification. The language used in the extension doc is pretty confusion on this IMO, I think it's fair that someone would interpret "transfers ownership" as meaning the driver is now responsible of closing it.

yshui:
I ran the attached example and it failed with `DeviceLost` on line 188. 

Driver is AMDGPU. Since this involves both vulkano and glium I am not sure where to report the bug?

@fayalalebrun

fayalalebrun:
@yshui Could you try the `gl-interop` example in the vulkano repository? The example included here is out of date.

PR #1968, opened 2021-11-20
Add gl_context method to Headless

fayalalebrun:
At the moment there is an inconsistency between the `Display` and `Headless` glutin backends. While `Display` has a `gl_window` method allowing you to get a reference to the inner glutin context, `Headless` has no such equivalent.

This adds the `gl_context` method to `Headless`, which serves the same function as the `gl_window` method in `Display`.

Such functionality can be very useful if for example you are interfacing with a third party library which requires detailed information about the OpenGL context.

### DIFF
--- a/src/backend/glutin/headless.rs
+++ b/src/backend/glutin/headless.rs
@@ -5,7 +5,7 @@ use crate::debug;
 use crate::context;
 use crate::backend::{self, Backend};
 use std::rc::Rc;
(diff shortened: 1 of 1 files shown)


est31:
Hmm I'm wondering about the exposing of the `Takeable` type in the public API. Not sure if it's even needed, and it is mostly an implementation detail. Will take a look if it can be replaced.

PR #1954, opened 2021-07-23
Restructure book files for mdbook

NifleySnifley:
- Moved markdown files for the book to `book/src/`
- Moved image files for the book to `book/src/resources/`
- Moved `book/README.md` to `book/src/` and used it as the introduction for the mdbook
- Added configuration files and `.gitignore` for mdbook

PR #1970, opened 2021-11-26
Make the tutorial easier to navigate.

Peter-Crawley:
I was using the tutorial and I found it annoying to navigate between the pages, here is a pull request to resolve this.

Changes made:
- Moved all images to book/assets/
- Added previous and next navigation links to each page
- Combined README and SUMMERY


est31:
Thanks but I prefer to go with mdbook instead, which seems kinda standard, see #1954.

Peter-Crawley:
Maybe consider merging this as a temporary measure until #1954 is ready, it should not cause any issues.

PR #1973, opened 2021-12-11
Switch from takeable-option to Option

est31:
This uses more commonly known types.


est31:
Nope, this won't fly. `takeable_option` is here to stay (for now).

PR #1977, opened 2021-12-20
Add context trait for driver & devices UUID retrieval

fayalalebrun:
When sharing resources between OpenGL and another API steps must be taken to ensure that the driver and device being used is the same on both sides.  This is especially important on systems with more than one graphics card, or where drivers may be selected on a per-program basis (As is the case on Linux).

For this purpose `EXT_external_objects` adds the possibility to get the UUID's of the driver and devices being used by the OpenGL driver. This can used, for example, in a Vulkan application which intends to share resources with the GL context, to select a compatible physical device.

### DIFF
--- a/src/context/uuid.rs
+++ b/src/context/uuid.rs
@@ -3,7 +3,7 @@ use super::Context;
 /// Describes an error preventing the retrieval of the uuid.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum UuidError {
-     /// EXT_external_objects is not supported by the driver
+    /// EXT_external_objects is not supported by the driver
     ExtensionNotPresent,
 }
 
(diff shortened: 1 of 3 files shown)


src/context/uuid.rs
```diff
@@ -0,0 +1,57 @@
...
+
+	f.write_str(desc)
+    }
+}
+
+impl std::error::Error for UuidError {}
+
+
+
+impl ContextUuidExt for Context {
```
est31, line 26: This trait is only implemented once, right? Why is it an extra trait and why is it not separate functions on the `Context` struct?

fayalalebrun: Yes, it's only implemented once and probably will never be implemented by any other struct. 

Since most glium users likely won't be using this functionality, I thought it would be a good idea to compartmentalize it and make it opt-in via the use of an extension trait. Of course the drawbacks are the added verbosity in the library as well as some user programs having to use an additional line `use glium::ContextUuidExt`.

If you think the drawbacks outweigh the benefits, let me know and I can make the methods part of the `Context` struct.

est31: I think it's better they become parts of the `Context` struct. `Context` already has a ton of functions, these two won't add too much noise. The impl block can stay in this file.

src/context/uuid.rs
```diff
@@ -35,7 +31,9 @@ impl ContextUuidExt for Context {
 	Ok(data)
     }
 
-    fn device_uuids(&self) -> Result<Vec<[u8; 16]>, Self::Error> {
+    /// Returns the UUIDs of the devices being used by this context. Useful to ensure
```
est31, line 34: Can you put this onto two lines with one empty line between? So that the first sentence is shown in the preview and the rest isn't.

src/context/uuid.rs
```diff
@@ -21,12 +19,10 @@ impl std::fmt::Display for UuidError {
...
 impl std::error::Error for UuidError {}
 
-
-
-impl ContextUuidExt for Context {
-    type Error = UuidError;
-
-    fn driver_uuid(&self) -> Result<[u8; 16], Self::Error> {
+impl Context {
+    /// Returns the UUID of the driver currently being used by this context. Useful to
```
est31, line 23: same here

src/context/uuid.rs
```diff
@@ -0,0 +1,57 @@
...
+impl std::error::Error for UuidError {}
+
+impl Context {
+    /// Returns the UUID of the driver currently being used by this context. Useful to
+    /// ensure compatibility when sharing resources with an external API.
+    pub fn driver_uuid(&self) -> Result<[u8; 16], UuidError> {
+        if !self.extensions.gl_ext_semaphore && !self.extensions.gl_ext_memory_object {
+	    return Err(UuidError::ExtensionNotPresent)
+	}
+	let mut data = [0u8;16];
```
est31, line 29: missing space

src/context/uuid.rs
```diff
@@ -0,0 +1,57 @@
...
+	unsafe {self.gl.GetUnsignedBytevEXT(crate::gl::DRIVER_UUID_EXT, data.as_mut_ptr())};
```
est31, line 30: Can you add spaces after the { and before the }

src/context/uuid.rs
```diff
@@ -0,0 +1,57 @@
...
+	}
+	let mut n = std::mem::MaybeUninit::<i32>::uninit();
+	let n = unsafe {
+	    self.gl.GetIntegerv(crate::gl::NUM_DEVICE_UUIDS_EXT, n.as_mut_ptr());
+	    n.assume_init()
+	};
+
+	let mut res = Vec::with_capacity(n as usize);
+	for i in 0..n {
+	    let mut data = [0u8;16];
```
est31, line 48: missing space

src/context/uuid.rs
```diff
@@ -0,0 +1,57 @@
...
+	    unsafe {self.gl.GetUnsignedBytei_vEXT(crate::gl::DEVICE_UUID_EXT, i as u32, data.as_mut_ptr())};
```
est31, line 49: Can you add spaces after the { and before the }

est31 (changes requested):

src/context/uuid.rs
```diff
@@ -13,45 +13,55 @@ impl std::fmt::Display for UuidError {
...
-	Ok(res)
-    }
-}
+        if !self.extensions.gl_ext_semaphore && !self.extensions.gl_ext_memory_object {
+            return Err(UuidError::ExtensionNotPresent);
+        }
+        let mut n = std::mem::MaybeUninit::<i32>::uninit();
+        let n = unsafe {
+            self.gl
+                .GetIntegerv(crate::gl::NUM_DEVICE_UUIDS_EXT, n.as_mut_ptr());
```
est31, line 48: This is not needed. The stuff inside unsafe can be on the same line.

src/context/uuid.rs
```diff
@@ -13,45 +13,55 @@ impl std::fmt::Display for UuidError {
...
-	}
-	let mut data = [0u8;16];
-	unsafe {self.gl.GetUnsignedBytevEXT(crate::gl::DRIVER_UUID_EXT, data.as_mut_ptr())};
-	Ok(data)
+            return Err(UuidError::ExtensionNotPresent);
+        }
+        let mut data = [0u8; 16];
+        unsafe {
+            self.gl
+                .GetUnsignedBytevEXT(crate::gl::DRIVER_UUID_EXT, data.as_mut_ptr())
```
est31, line 33: This is not needed. The stuff inside unsafe can be on the same line.

est31 (changes requested):

est31 (approved):
Thanks!

PR #1984, opened 2022-01-17
Added link for unexplained topic in the book.

ayushk7:
Stack overflow link added for why we need to multiply normal with the transpose of the inverse of transformation matrix. Makes easy for beginners to navigate to the problem.

### DIFF
--- a/book/tuto-08-gouraud.md
+++ b/book/tuto-08-gouraud.md
@@ -64,7 +64,7 @@ void main() {
 
 We also need to multiply the normal by the matrix, but the transformations are a bit different
 and the calculation a bit weird. Since I didn't go into details about how matrices work,
-I won't go into details about why you have to use the transpose of the inverse.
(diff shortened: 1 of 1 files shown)


book/tuto-08-gouraud.md
```diff
@@ -64,7 +64,7 @@ void main() {
 
 We also need to multiply the normal by the matrix, but the transformations are a bit different
 and the calculation a bit weird. Since I didn't go into details about how matrices work,
-I won't go into details about why you have to use the transpose of the inverse.
+I won't go into details about why you have to use the transpose of the inverse ([click for details](https://stackoverflow.com/questions/13654401/why-transform-normals-with-the-transpose-of-the-inverse-of-the-modelview-matrix)).
```
est31, line 67:
```suggestion
I won't go into details about why you have to use the transpose of the inverse ([click for details](https://stackoverflow.com/q/13654401)).
```

est31 (changes requested):

est31 (approved):
Thanks!

PR #1989, opened 2022-02-03
Fix texture states not being updated in texture::any::new_from_fd

yshui:


est31:
@fayalalebrun as you added the code in #1965, does this look good?

fayalalebrun:
@est31 Yes, this is a valid and needed fix.

est31 (approved):
Approved per @fayalalebrun 's review

PR #1993, opened 2022-02-08
sRGB clear_color if statement logic fix

eligum:
Pretty simple change, now if `color_srgb` is false it will properly disable the `gl:FRAMEBUFFER_SRGB` option. There was a mistake in the if statement logic.


est31 (approved):
Thanks!

Issue #1791, opened 2019-09-24
Procedural macro for Vertex trait


quebin31:
Is there any reason because Vertex trait cannot be implemented using a procedural macro? Resulting code would be more idiomatic:

**Preview:**
```rust
#[derive(Copy, Clone, Vertex)]
struct Vertex3 {
    position: [f32; 3],
    tex_coords: [f32; 2],
}
```

est31:
Good point! In fact, proc macros would be an awesome idea. Historically, glium used to have a plugin that used to provide such functionality, see the macro directory in the source code (at this point, it should be removed I guess). Also, it's mentioned in the limitations section of the [README](https://github.com/glium/glium/blob/master/README.md):

> Rust plugins are not stable, so you will have to use macros such as `implement_vertex!(MyStruct)` instead of `#[derive(GliumVertex)]`.

Note that both are very old and predate the stabilization of derive macros.

I've wondered about possibly creating a derive macro myself but discarded the idea because glium has a large compile time and compiling a derive macro as well would increase it even further. However, ultimately it's up to the users so we could create a glium_derive crate that users could opt into.

quebin31:
Yeah, creating another crate `glium_derive` and just using a feature would do the trick.

```toml
[dependencies]
glium = { version = "*", feature = ["derive_macros"] }
```

est31:
PRs etc. to do this are definitely welcome. Note that the macros crate has multiple macros so maybe they could serve as inspiration.

quebin31:
`macros` crate inside `glium` isn't being used right? Changes proposed here should be done there?

est31:
It's not used. But I think it's easier to start over. There seems to be a derive impl at https://github.com/voidxnull/glium_derive but it's abandoned. Maybe we can ask @voidxnull to transfer the repo to the glium org.

voidxnull:
@est31 I'll be happy to transfer the repo since I don't use glium anymore.

est31:
@tomaka  would it be possible to facilitate such a transfer?

tomaka:
What should I do? You don't need the approval of the destination to transfer a repo.

voidxnull:
@tomaka This is what I get when I try to transfer the repo: `You don’t have the permission to create public repositories on glium`

voidxnull:
Still can't transfer, it fails with the same error.

tomaka:
It was disabled in the settings! It should work now.

voidxnull:
[...]

PR #1994, opened 2022-02-12
Dropped glium-macros crate

Cerber-Ursi:
This is an intermediate change, more technical then meaningful.

This PR allows the glium_macros crate to compile on current stable Rust. It exports two derive macros, corresponding to the `Vertex` and `Uniform` traits in `glium`. The implementation for `Vertex` is exactly the same as in the previously available code; the implementation for `Uniform` was changed according to the change in the trait itself.

Note that the derived implementations are _not_ tested for correctness yet (and they will surely change, since at least one part of them currently seems to be unsound); the only test currently present is a "smoke" test, showing that the derive can be compiled. This will be changed with further PRs.


est31:
Thanks, but note that this effort has already happened in the form of the `glium_derive` crate: https://github.com/glium/glium_derive . See also the discussion in #1791. Maybe focus your efforts there?

Ideally the macros crate would be removed from this repo. Feel free to update your PR to do this.

Cerber-Ursi:
Thanks, I've missed it, will look through there.

est31:
This is partially the fault of the project for not removing that crate for good.

Cerber-Ursi:
Sorry for the delay. Updated.

PR #1997, opened 2022-02-20
Make `Display::new` take a `&EventLoopWindowTarget`

### LINKED ISSUE #1996: No way to dynamically open new windows?
Currently `Display::new` takes a `&EventLoop` instead of a `&EventLoopWindowTarget`. It seems like this means that there is no way to create new `Display`'s after `EventLoop::run` is called, and thus there is no way to dynamically open new windows. Since `Display::new` only needs the event_loop to call `ContextBuilder::build_windowed` it should be possible for `Display::new` to take a `&EventLoopWindowTarget`.

birktj:
This PR changes `Display` new to take a `&EventLoopWindowTarget` instead of  a `&EventLoop`. Since `EventLoop`  derefs to `EventLoopWindowTarget` this should not be a breaking change. This change allows one to dynamically create new windows.

I have also added a small example to show how to dynamically create new windows.

This fixes #1996

### DIFF
--- a/examples/multiple_windows.rs
+++ b/examples/multiple_windows.rs
@@ -1,0 +1,69 @@
+use glium::glutin::{window::WindowBuilder, ContextBuilder};
+use glium::glutin::event_loop::{EventLoop, ControlFlow};
+use glium::glutin::event::{Event, WindowEvent, ElementState, MouseButton};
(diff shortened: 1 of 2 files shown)


est31 (approved):
Thanks!

PR #2002, opened 2022-04-15
Remove trivial and unnecessary UB

digama0:
I don't understand why it was written this way. [glGenTextures](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGenTextures.xhtml) is passed a mutable pointer which is set as an out parameter, but the existing code was transmuting a shared reference to the buffer to a mutable pointer, which is undefined behavior. The fix is trivial: just make the buffer mutable and pass a mutable reference instead.


est31 (approved):
Thanks!

PR #2005, opened 2022-05-08
Book 13 Blinn-phong shading image link is broken.

peter5600:
Link should be resources/tuto-13-specular.png instead its resources/resources/tuto-13-specular.png
Tested on website and it fixes the issue.


est31 (approved):
Thanks!

Issue #361, opened 2015-01-24
Implement image load/store


tomaka:

tomaka:
Depends on #945

tomaka:
Requires implementing memory barriers for textures as well.

dylanede:
What's the status of this issue now? What can I do to help?

darkstalker:
Stepped into this while trying to use compute shaders. Can't pass a `Texture2d` into a `uniform image2D`, so I can't write to a texture from a compute shader. I think it's about calling `glBindImageTexture`, but duno about the glium implementation details.

dylanede:
@darkstalker I think the main thing that needs work for this to be implemented is as tomaka mentioned before - tracking of memory access for textures is required, with memory barriers introduced as and when necessary. Presumably memory barriers need to be applied for any textures possibly written to by a shader using image load/store, just before the next use of the texture. Is that the approach you recommend @tomaka? I can have a stab at implementing that if that's OK with you.

Tsarpf:
Hit the same thing when trying to use compute shaders. I guess I'll try to use a custom buffer and then convert that back to a texture.

IgrauBezRuk:
pizdec. Seems like it still not sloved, right?

Boscop:
Any update on this? :)

Lucky4Luuk:
So...

PR #2006, opened 2022-05-14
Add support for Image Load/Store

fayalalebrun:
This MR adds support for Image Load/Store functionality. This is a continuation of the work started in #1691 and solves #361.

- Automatic synchronization is achieved the same way as it was previously done with buffers. A counter with the draw/invocation id is kept, and each texture stores the draw id when it was last written to in a shader. If the texture is used whether that is as a sampler uniform, an image unit, or a framebuffer object, the last time the appropriate memory barrier was issued is checked, and if it is before the texture was written to, an appropriate memory barrier is issued.

Due to the current implementation of framebuffer objects in glium, it is rather difficult to perform proper synchronization, since attachments are converted to `RawAttachment` and this loses its reference to the original texture.

Thus, synchronization with framebuffers is only done whenever a texture's `as_surface` function is called. Additionally, a function called `sync_shader_writes_for_surface` is added, which can be used to perform synchronization for framebuffers in more complex situations.

- `ImageUnit`'s parameters are validated in runtime. Additionally, a format is required upfront, since directly converting the texture's format to ImageUnit's format seems to be too complex. However, compatibility checks are done upfront, according to the specification. 

- A new example is added, which showcases a reasonably complex use case for the new functionality.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ backtrace = "0.3.2"
 lazy_static = "1.0"
 smallvec = "1.0"
 fnv = "1.0.5"
+bitflags = "1.0"
 
 [build-dependencies]
 gl_generator = "0.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,6 @@ backtrace = "0.3.2"
 lazy_static = "1.0"
 smallvec = "1.0"
 fnv = "1.0.5"
-bitflags = "1.0"
 
 [build-dependencies]
 gl_generator = "0.14"
--- a/examples/compute_image.rs
+++ b/examples/compute_image.rs
@@ -1,0 +1,211 @@
+// Based on https://github.com/Erkaman/image-load-store-demo
+//
+// This example showcases image load/store functionality in compute shaders, which
+// allows you to read and write to arbitrary textures in a shader.
+//
+// Three compute shaders are launched. The first one draws the mandelbrot fractal. The second
+// one applies a box filter on the output of the first. Finally, the third one copies data
+// the original unsigned integer texture to a signed normalized integer texture.
+// Finally, this last texture is blitted to the display framebuffer.
+
+#[macro_use]
+extern crate glium;
+
+use std::time::Instant;
+mod support;
+
+use glium::{texture::UnsignedTexture2d, uniform, Surface, Texture2d};
+
+fn main() {
+    // building the display, ie. the main object
+    let event_loop = glutin::event_loop::EventLoop::new();
+    let wb = glutin::window::WindowBuilder::new();
+    let cb = glutin::ContextBuilder::new()
+        .with_vsync(true)
+        .with_gl(glutin::GlRequest::Latest);
+    let display = glium::Display::new(wb, cb, &event_loop).unwrap();
+
+    let start_time = Instant::now();
+
+    let fract_texture = UnsignedTexture2d::empty_with_format(
+        &display,
+        glium::texture::UncompressedUintFormat::U8U8U8U8,
+        glium::texture::MipmapsOption::NoMipmap,
+        1024,
+        1024,
+    )
+    .unwrap();
+
+    let final_texture = Texture2d::empty_with_format(
+        &display,
+        glium::texture::UncompressedFloatFormat::U8U8U8U8,
+        glium::texture::MipmapsOption::NoMipmap,
+        1024,
+        1024,
+    )
+    .unwrap();
+
+    let fractal_shader = glium::program::ComputeShader::from_source(&display, r#"\
+#version 430
+layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;
+
+
+
+uniform uint uWidth;
+uniform uint uHeight;
+uniform float uTime;
+uniform layout(binding=3, rgba8ui) writeonly uimage2D uFractalTexture;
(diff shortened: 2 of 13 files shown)


src/uniforms/image_unit.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    /// Set the mip level that will be bound
+    pub fn set_level(mut self, level: u32) -> Result<Self, ImageUnitError> {
+        self.0.mipmap(level).ok_or(ImageUnitError::NoMipmapAtLevel(level))?;
+        self.1.level = level;
+        Ok(self)
+    }
+
+    /// Sets the layer of the texture to bind, or None to disable layer binding
+    /// TODO: only implement this for texture types where layering makes sense
```
est31, line 88: Can you move the TODO out of the doc comment?

src/uniforms/image_unit.rs
```diff
@@ -0,0 +1,335 @@
...
+    R8snorm,
+}
+
+impl ImageUnitFormat {
+    fn get_total_bits(&self) -> usize {
+        match self {
+            ImageUnitFormat::RGBA32F => 4*32,
+            ImageUnitFormat::RGBA16F => 2*32,
+            ImageUnitFormat::RG32F => 2*32,
+            ImageUnitFormat::RG16F => 1*32,
```
est31, line 241: This should be 2*16. Same result, but different thning. Same applies to the other places this occurs in the match.

src/texture/buffer_texture.rs
```diff
@@ -550,6 +550,11 @@ impl<'a> TextureExt for BufferTextureRef<'a> {
         unsafe { ctxt.gl.BindTexture(gl::TEXTURE_BUFFER, self.texture); }
         gl::TEXTURE_BUFFER
     }
+
+    fn prepare_for_access(&self, _: &mut CommandContext<'_>, access_type: crate::TextureAccess) {
+        // TODO: Right now this type of texture cannot be used in an image unit
```
est31, line 555: I wonder if a panic or some other error would be better here?

src/program/raw.rs
```diff
@@ -615,6 +617,8 @@ impl RawProgram {
             fence.insert(&mut ctxt);
         }
 
+	ctxt.state.next_draw_call_id += 1;
```
est31, line 620: indentation doesn't line up.

src/program/raw.rs
```diff
@@ -572,6 +572,8 @@ impl RawProgram {
...
```
est31, line 575: indentation doesn't line up.

src/lib.rs
```diff
@@ -1143,10 +1160,13 @@ impl fmt::Display for DrawError {
...
                 "Restarting indices (multiple objects per draw call) is not supported by the backend",
             ClipPlaneIndexOutOfBounds =>
-                "Tried to enable a clip plane that does not exist."
+                "Tried to enable a clip plane that does not exist.",
+            InsufficientImageUnits =>
+                "Tried to use more image uniforms that the implementation has support for",
         };
         match self {
             UniformTypeMismatch { name, expected } =>
+
```
est31, line 1169: Can you remove this blank line?

src/lib.rs
```diff
@@ -377,6 +377,20 @@ trait TextureExt {
...
+    /// with image load/store
+    fn prepare_for_access(&self, _: &mut CommandContext<'_>, access_type: TextureAccess);
+}
+
+/// Ways a texture could possibly be accessed after being written to in a shader via an image unit
+enum TextureAccess {
+    /// Regular texture fetches within shaders
+    TextureFetch,
+    /// Texture being used as an image unit inside of a shader
+    ImageUnit{ will_write: bool},
```
est31, line 391: Can this use standard Rust formatting?

examples/compute_image.rs
```diff
@@ -0,0 +1,211 @@
+// Based on https://github.com/Erkaman/image-load-store-demo
```
est31, line 1: Maybe due to the license it would be best to copy the MIT license here?

Cargo.toml
```diff
@@ -44,6 +44,7 @@ backtrace = "0.3.2"
 lazy_static = "1.0"
 smallvec = "1.0"
 fnv = "1.0.5"
+bitflags = "1.0"
```
est31, line 47: is the bitflags crate still used somwhere? It can be removed now, right?

est31 (changes requested):
I've given it a cursory (non-deep) review as I'm not very familiar with that part of OpenGL. If there are two PRs for it, it seems to be a wanted feature though.

est31:
As for the CI failure, you can rebase this PR, then it should be resolved.

src/uniforms/image_unit.rs
fayalalebrun: I wrote it this way to reflect the way it is done in the API spec. I've changed it to match your preference, however.

src/texture/buffer_texture.rs
fayalalebrun: I think it is better to leave this empty, as this type cannot be made into an image unit at the moment, therefore the missing implementation has no adverse effect.

If we added a panic here it would make it impossible to use this type as a shader uniform, as it would always result in a panic.

est31 (approved):
Thanks!

PR #2013, opened 2022-06-09
Added support for location binding in 'implement_vertex'

jorundev:
Firstly, let me tell you
- I don't know glium very much
- I never contributed to a project like that before
- My understanding of OpenGL is basic
- My understanding of rust is.. okay-ish
- My first language is not english. I'm good at it but I can struggle when it gets technical

The code I provided can't be merged right now because of.. reasons. But I can't really do anything more before discussing the topic with you first

I wanted to make GLSL exported from spirv-cross work with glium.  spirv-cross generates GLSL like this
```glsl
layout(location = 0) in vec2 input_position;
layout(location = 1) in vec3 input_color;
```

usually, in raw Opengl, that enough for me. I do some glAttribPointer magic and it's alright

But here comes the glium documentation for [glium::implement_vertex](https://docs.rs/glium/latest/glium/macro.implement_vertex.html#):

> When it comes to using to using your vertex array in a shader you must make sure that all your attribute variables match the field names in the struct you are calling calling this macro for.
> 
> So, if you have a vertex_position atribute/input in your shader, a field named vertex_position must be present in the struct. Ohterwise the drawing functions will panic.

oh no! if spirv-cross (or naga, or anything really) generates something inelegant, i'll have to adapt my struct to it, or modify the generated GLSL somehow. I didn't like this

There might be a solution for this problem that I didn't think of. If that's the case, tell me so I can open an issue and we can sort this out there. But I decided to dive into glium's source code to see if I could make something work

I came with this **VERY INCOMPLETE** implementation. The macro I made works, but is inelegant. The implement_vertex macro has a 'normalize' option thing so I used the same logic to make a 'location' option. But it's not possible to mix 'normalize' and 'location' together, which is a shame. This is a problem of implement_vertex in the first place, but I didn't want to change too much since I don't know the source code enough

I might be missing a couple of checks too. I will need the advice of someone who knows the stuff© to continue.

Here's how to use a location in my implementation

```rust
#[derive(Copy, Clone)]
pub struct Vertex {
    pub position: [f32; 2],
    pub color: [f32; 3],
}

glium::implement_vertex!(Vertex, position location(0), color location(1));

```

This works for me (on a simple triangle with attribute names that do not match). glium still panics if the vertex format is not right (if you made a mistake in the locations for example)

### DIFF
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -286,6 +286,18 @@ impl VertexArrayObject {
             }
         }
 
+        // checking for duplicate attribute locations
+        for &(_, ref bindings, _, _, _) in vertex_buffers {
+            for i in 0..bindings.len() {
+                for o in 0..bindings.len() {
+                    if i != o && bindings[i].2 == bindings[o].2 {
+                        panic!("The program attribute `{}` has the same binding location as program attribute `{}` (binding location {})",
+                               bindings[i].0, bindings[o].0, bindings[i].2)
+                    }
+                }
+            }
+        }
+
         // checking for missing attributes
         for (&ref name, attribute) in program.attributes() {
             let mut found = false;
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -262,7 +262,7 @@ impl VertexArrayObject {
             for &(ref name, _, location, ty, _) in bindings.iter() {
                 let attribute = match location {
                     -1 => {
-                        // No location precised in Vertex Format. Check name instead
+                        // No location specified in Vertex Format. Check name instead
                         match program.get_attribute(Borrow::<str>::borrow(name)) {
                             Some(a) => a,
                             None => continue,
@@ -549,7 +549,7 @@ unsafe fn bind_attribute(ctxt: &mut CommandContext<'_>, program: &Program,
 
         let attribute = match location {
             -1 => {
-                // No location precised in Vertex Format. Check name instead
+                // No location specified in Vertex Format. Check name instead
                 match program.get_attribute(Borrow::<str>::borrow(name)) {
                     Some(a) => a,
                     None => continue,
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -288,11 +288,11 @@ impl VertexArrayObject {
 
         // checking for duplicate attribute locations
         for &(_, ref bindings, _, _, _) in vertex_buffers {
-            for i in 0..bindings.len() {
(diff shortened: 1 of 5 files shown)


jorundev:
I added checks for duplicate attribute locations and modified the documentation. I'm not sure if I need to add a check for attribute locations specified with the macro but not present in the program attributes

src/vertex_array_object.rs
```diff
@@ -286,6 +286,18 @@ impl VertexArrayObject {
             }
         }
 
+        // checking for duplicate attribute locations
+        for &(_, ref bindings, _, _, _) in vertex_buffers {
+            for i in 0..bindings.len() {
```
est31, line 291:
```suggestion
            for (i, bi) in bindings.iter().enumerate() {
```
you can check iterators here, then you don't need to index below

jorundev: true!

est31 (approved):
Thanks!

jorundev:
The next step is changing the implement_vertex macro so it can accept multiple options. Then the PR will be complete

src/vertex_array_object.rs
```diff
@@ -520,12 +544,24 @@ unsafe fn bind_attribute(ctxt: &mut CommandContext<'_>, program: &Program,
...
-    for &(ref name, offset, ty, normalize) in bindings.iter() {
+    for &(ref name, offset, location, ty, normalize) in bindings.iter() {
         let (data_type, elements_count, instances_count) = vertex_binding_type_to_gl(ty);
 
-        let attribute = match program.get_attribute(Borrow::<str>::borrow(name)) {
-            Some(a) => a,
-            None => continue
+        let attribute = match location {
+            -1 => {
+                // No location precised in Vertex Format. Check name instead
```
est31, line 552: I don't think "precised" is a valid english word.

jorundev: you are right

jorundev:
I don't see how to add support for multiple attribute options without creating a mess. This looks weird to me:
```rust
glium::implement_vertex!(Vertex2D, position location(0) normalize(true), color location(1));
```

I think it would make more sense written like this:
```rust
#[derive(Copy, Clone, glium::vertex::Vertex)]
pub struct Vertex2D {
    #[normalize]
    #[location(0)]
    pub position: [f32; 2],

    #[location(1)]
    pub color: [f32; 3],
}
```

This would be a pretty significant change (and a breaking one). Should I open an issue for it @est31 ? And what should I do with this PR ?

est31 (approved):
Thanks!

PR #2025, opened 2022-11-02
Update glutin to 0.30

est31:
https://github.com/rust-windowing/glutin/releases/tag/v0.30.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,13 +27,19 @@ exclude = ["doc"]
 edition = "2021"
 
 [features]
-default = ["glutin"]
+default = ["glutin_backend"]
+glutin_backend = ["glutin", "winit"]
 unstable = [] # used for benchmarks
 test_headless = []  # used for testing headless display
 vk_interop = [] # used for texture import from Vulkan
 
 [dependencies.glutin]
-version = "0.29"
+version = "0.30"
+features = []
+optional = true
+
+[dependencies.winit]
+version = "0.27"
 features = []
 optional = true
 
--- a/book/tuto-01-getting-started.md
+++ b/book/tuto-01-getting-started.md
@@ -48,8 +48,8 @@ Initializing an OpenGL window with glutin can be done using the following steps:
 fn main() {
     use glium::glutin;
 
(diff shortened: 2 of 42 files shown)


kchibisov:
@est31 if you need help with glutin update you can ping me on #winit IRC/matrix or here. In general I think you can remove winit entirely from the glutin integration? Like some functions like rebuilding the context could be done without a window for example, and you may only need `raw-window-handle` in some places.

If you still want `winit` for outside the examples you might want to look at [`glutin-winit`](https://github.com/rust-windowing/glutin/pull/1517). Though, I don't see a need for winit inside your code.

The window and context are also not related to each other since on Android you have your window live for a cycle of `Resume` and `Suspend` events, but `Context` lives as long as it wants.

Though given that you have a signature of `swap_buffers(&self)` it seems like you may want to either change the `Backend` trait or tie `GlSurface` and `GlContext` from new glutin together. And have a method to change surface associated with the context.

You can also consider using Traits glutin exposes, so you can have it automatically working not only for top-level stuff, but for `api::egl::Surface`, since they all implement the same traits, but it's not really an issue, users can always wrap their specialized stuff into `glutin::surface::Surface`.

est31:
@kchibisov thanks for reaching out and offering help with the migration. I will join IRC to discuss it with you.

est31:
@wez thanks for the feedback. I need to find time to get back to this PR, right now and at least the coming week I don't have much time for it. If you want, help is welcome :). My goal is to have glium use winit by default, but to make it optional.

Melchizedek6809:
@est31 I've been working a bit starting from your branch and got a triangle to show, still super rough and dirty but might be helpful: https://github.com/Melchizedek6809/glium/tree/update_glutin
If you're still busy I can continue working on this, kinda fun to dive into the internals, just want to make sure that there's no wasted effort.

est31:
@Melchizedek6809 that's very cool! I probably won't have time to dedicate on this soon-ish so it would be really great if you took over. It's great that you got a triangle to render :).

Melchizedek6809:
@est31 Alright, then I'll continue and ask you once there's something sensible to review :)

est31:
Closing in favour of #2036

PR #2026, opened 2022-11-03
Fixed using of gl_PointSize on GLES contexts

Melchizedek6809:
Good evening, first off, thank you for maintaining this brilliant library! It's been a pleasure to use.

Apart from that I've stumbled upon a problem with the handling of `GL_PROGRAM_POINT_SIZE` in ES contexts, there is a check which throws an error if the context is not GL 3.0 or higher, however `GL_PROGRAM_POINT_SIZE` support seems to be available from 2.0 onwards. On ES that enum is not available and gl_PointSize seems to be available by default.

Because of this I've removed the version check and also added changes so that the glEnable/Disable calls for GL_PROGRAM_POINT_SIZE will only occur on GL contexts where they are necessary, and not in ES contexts since they result in a panic there.

With this patch applied my program at least runs great on both GL and ES contexts.

1. https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_PointSize.xhtml - gl_PointSize is available since GLSL 1.1
2. https://registry.khronos.org/OpenGL-Refpages/es3/html/glEnable.xhtml - `GL_PROGRAM_POINT_SIZE` not in the list

### DIFF
--- a/src/program/program.rs
+++ b/src/program/program.rs
@@ -82,7 +82,7 @@ impl Program {
                     return Err(ProgramCreationError::TransformFeedbackNotSupported);
                 }
 
-                if uses_point_size && !(facade.get_context().get_version() >= &Version(Api::Gl, 3, 0)) {
+                if uses_point_size && (facade.get_context().get_version().0 == Api::Gl) && !(facade.get_context().get_version() >= &Version(Api::Gl, 3, 0)) {
                     return Err(ProgramCreationError::PointSizeNotSupported);
                 }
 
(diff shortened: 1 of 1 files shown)


est31:
Can you change it so that the version check for non-openGL ES is still around? The issue is, we still support older OpenGL versions, so still need to check if extensions are supported *if* the version is older than the one that introduced some extension as mandatory.

Melchizedek6809:
Alright, changed it so that for non-ES contexts nothing changed.

What do you think about lowering the version requirement for gl_PointSize from 3.0 to 2.0 though?
Since according to the Khronos docs it's been around since GLSL 1.1.

Melchizedek6809:
Hey there, anything else you'd like me to change?

est31:
Thanks for the PR! Sorry for the delay.

> What do you think about lowering the version requirement for gl_PointSize from 3.0 to 2.0 though?

I'm not entirely certain. First, I don't know what exactly the minimum supported OpenGL version of glium is (i tried to look it up but couldn't find anything explicit). Second, `gl_PointSize` can be used both in vertex and geometry shaders, and it requires OpenGL 3.2 to work in geometry shaders. It seems that there is only a single flag that gets activated when either vertex or geometry shaders use `GL_PROGRAM_POINT_SIZE`, which means we can't distinguish and have to assume the worst case. With that in mind, the check is actually too tolerant. It might make sense to split it up into two, but don't know how many users of < OpenGL 3.0 there still are, whether it's worth it. But feel free to make a PR.

The PR, a fix for ES, is good.

est31 (approved):

PR #2029, opened 2022-11-21
Calculate vertex stride based on bindings size

### LINKED ISSUE #2028: `vertex_array_object.rs` VertexAttribPointer stride being set incorrectly
So in my project, I have a mesh file that has the data already ready to be sent to the GPU. I create bindings dynamically, and call `VertexBuffer::new_raw`. Then, when VertexAttribPointer in  vertex_array_object.rs is called, the stride parameter is incorrect.

Example:

Vertex attributes (Generated)
```
 [("position", 0, 0, F16F16F16, false), ("texturelocation", 6, 1, F16F16, false)]
```
So here we see the vertex consists of 3x F16 for position, and 2x F16 for texture location. This means that each vertex is 5x2 bytes long (10 bytes).

So I call `VertexBuffer::new_raw` like so:
```rust
VertexBuffer::new_raw(
                display,
                vertex_buffer.as_slice(),
                Cow::Owned(bindings),
                10,
            )
```

Note that `vertex_buffer` is simply a array of u8 bytes from the file being read into the GPU directly.

but then, when rendering, the vertex size is ignored and set to 1 instead!

## Call (vertex_array_object.rs)
```
ctxt.gl.VertexAttribPointer(1, 2, 5131, 1, 6,)
```

Note how the 4th parameter here is 1 in the call withint glium? This should be 10 as it is what I set in my vertex stride size.

rnd-ash:
Potentially a breaking change. This needs more testing.

This fixes #2028

VertexAttribPointer calls now take into account the total size of all bindings for the vertex in order to better calculate the vertex stride size

### DIFF
--- a/src/vertex_array_object.rs
+++ b/src/vertex_array_object.rs
@@ -543,7 +543,10 @@ unsafe fn bind_attribute(ctxt: &mut CommandContext<'_>, program: &Program,
         ctxt.state.array_buffer_binding = vertex_buffer;
     }
 
-    // binding attributes
+    // Stride value taken from bindings
+    // Calculated based on the size of each vertex in bytes
+    let stride_bindings: usize = bindings.iter().map(|x| x.3.get_size_bytes()).sum();
+
     for &(ref name, offset, location, ty, normalize) in bindings.iter() {
         let (data_type, elements_count, instances_count) = vertex_binding_type_to_gl(ty);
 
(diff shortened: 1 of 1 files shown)


est31:
Right now the stride seems to be calculated from the `get_elements_size` function, which in most instances just calls `mem::size_of::<T>` for `T` being the vertex type. I'm not sure this is wrong, although the `new_raw` function you quoted in  #2028 does seem wrong indeed.

rnd-ash:
Re-opened as #2031. Accidently pushed a commit not intended for this change

PR #2031, opened 2022-11-21
(2) Calculate vertex stride based on bindings size (#2028) 

rnd-ash:


est31:
My comment from #2029 applies, I'm not sure the old code was wrong?

est31:
Closing as I got no reply.

PR #2032, opened 2022-12-08
Lowered version check for gl_PointSize from 3.0 to 2.0

Melchizedek6809:
Hello again,

this time I lowered the version requirement for using gl_PointSize from 3.0 to 2.0.
This should be sufficient since 3.2 / 4.0 is only required when used within a geometry / tesselation evaluation shader.
Since platform support is already checked when compiling those shaders we probably don't need to check so again afterwards.

Regarding the extensions that allow circumventing the version checks:
- gl_ext_geometry_shader4 : This seems to support it, and the new token PROGRAM_POINT_SIZE_EXT is just an alias for VERTEX_PROGRAM_POINT_SIZE

- gl_ext_geometry_shader : This seems to be ES specific, where gl_PointSize is always available

- gl_ext_tessellation_shader : This also seems to be ES specific, where gl_PointSize is always available

Apart from that if you prefer it I could change it and check for 2.0/3.2/4.0 depending on whether a geometry/tesselation shader is used in the program.

### DIFF
--- a/src/program/program.rs
+++ b/src/program/program.rs
@@ -82,7 +82,7 @@ impl Program {
                     return Err(ProgramCreationError::TransformFeedbackNotSupported);
                 }
 
-                if uses_point_size && (facade.get_context().get_version().0 == Api::Gl) && !(facade.get_context().get_version() >= &Version(Api::Gl, 3, 0)) {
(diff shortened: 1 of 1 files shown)


est31 (approved):
Looks good. It seems that the [minimum glsl version](https://registry.khronos.org/OpenGL-Refpages/gl4/html/gl_PointSize.xhtml) of `gl_PointSize` is available on tesselation and geometry shaders matches the minimum opengl versions of those shaders (at least when they became obligatory), so the table doesn't actually say much.

PR #2036, opened 2023-01-02
Update glutin to 0.30

Melchizedek6809:
Hey there,

I've been working on updating glium to work with glutin 0.30 and things are looking quite well, I've added a new example `triangle-new` because glutin needs to be initialized quite differently. I also restructured the overall example so that everything gets setup on receiving a `Resume` Event and dropped when a `Suspend` Event arrives, this apparently should only happen on Android, but I thought that examples should probably promote the most reliable/safe way of handling things. 

Apart from that I tried to keep changes to the API as little as possible, which is why I've added the `ContextSurfacePair` type. I've also experimented on another branch with just keeping the Context around, but that meant breaking the API since we can't make a Context current without a Surface, and we can't determine the size of the Framebuffer without a Surface.

Two other things where someone with more Rust experience may know of a better Solution:

1. Is there a better/more efficient way of turning a `&str` into a `&CStr`? I'm currently doing the following:
```rust
#[inline]
unsafe fn get_proc_address(&self, symbol: &str) -> *const c_void {
    let symbol = CString::new(symbol).unwrap();
    self.borrow().display().get_proc_address(&symbol) as *const _
}
```

2. I'm not sure about the `'static` lifetime for the `Display` struct in `src/backend/glutin/mod.rs:82`, I have a gut feeling that this is wrong, but I'm too new to Rust to know for certain. 

Oh and it currently only works on *nix Platforms, due to `glutin` not implementing `Surface::{width,height}` for CGL/WGL contexts which this implementation depends on, I'm looking into implementing that for glutin but wanted to create a draft request already since this might take some time.


est31:
> 1. Is there a better/more efficient way of turning a `&str` into a `&CStr`?

Sadly no. You have to allocate by creating a `CString` as the `&str` has no zero byte at the end, while CStr guarantees trailing zeroes.

> 2. I'm not sure about the `'static` lifetime for the `Display` struct

Not 100% sure, for that I have to see the error that happens if you remove it...

> Oh and it currently only works on *nix Platforms, due to `glutin` not implementing `Surface::{width,height}` for CGL/WGL contexts which this implementation depends on

Yeah if possible, getting the width/height on the surface should be cross platform.

Melchizedek6809:
> Sadly no. You have to allocate by creating a CString as the &str has no zero byte at the end, while CStr guarantees trailing zeroes.

Ah alright, too bad

> Not 100% sure, for that I have to see the error that happens if you remove it...

I get the following error without the static lifetime:
```rust
error[E0310]: the parameter type `T` may not live long enough
   --> src/backend/glutin/mod.rs:172:32
    |
172 |         let context = unsafe { context::Context::new(glutin_backend, checked, debug) }?;
    |                                ^^^^^^^^^^^^^^^^^^^^^ ...so that the type `T` will meet its required lifetime bounds
    |
help: consider adding an explicit lifetime bound...
    |
111 | impl<T: SurfaceTypeTrait + ResizeableSurface + 'static> Display<T> {
    |                                              +++++++++
```

> Yeah if possible, getting the width/height on the surface should be cross platform.

I've already opened a PR which implements the missing functionality on Windows/Mac: https://github.com/rust-windowing/glutin/pull/1551 So hopefully this won't be an issue for long.

est31:
> I get the following error without the static lifetime:

Yeah that's because `Display::new` requires 'static for its backend argument's  type. That's present because in that call we put the backend into a boxed trait object.

> I've already opened a PR which implements the missing functionality on Windows/Mac

Very cool :).

kchibisov:
I've released `glutin` `0.30.4` which has `width/height` implemented for every graphics backend, it still returns an `Option`, but it means that the call failed and not that you can do much(though, I'm not sure they could fail...). I'd assume you now don't really on winit anymore, given that you wanted stuff like that implicitly.

Be aware that on macOS context/width/height/resize calls will block off the main thread if the main thread is blocked, though I doubt many folks use thread per context anyway, given that before macOS was SIGILL-ing.

Melchizedek6809:
@est31 
Alright, now this branch depends on the version from crates.io and the new example seems to work on my devices at least.
Is there anything you'd like me to change?

est31:
@Melchizedek6809 I need to give this a closer look on the weekend. Do you have ideas what to do about the examples? I don't know yet which direction they should go so for now I'm only looking for ideas, not neccessarily having the migration implemented.

Melchizedek6809:
@est31
Maybe we could put the whole context creation/suspension/resumption handling into the example support module, with each example then only implementing a Trait with methods for initialization/drawing?

Since from what I can gather on first glance, most examples don't seem to differ in the way they create the Context or deal with the event loop, which might also lead to the specific examples not directly depending on glutin/winit anymore.

It might however make some of the smaller examples more verbose due to the state struct.

est31:
Sorry for coming back only now, my weekend was very busy. In general examples should be self-contained and small so that you know how to use the API from looking only at few lines of code. On the other hand, being able to focus only on the parts that matter is also very useful. I think for the very introductory examples it would be best to not use util modules but one could use it for more involved examples. It's a bit of an effort to port all of them.

Melchizedek6809:
@est31 
No problem, thank you for taking the time anyways.

Yeah, I though it might make some examples clearer if the whole context creation boilerplate gets moved somewhere else, since it barely changes, though as you metion it's probably bad for the tutorials and associated examples.

What do you think about using the same/similar structure but including it directly in the examples / tutorials so that it's not hidden away?

Yeah it's probably quite a bit of work to update all the examples and especially the tutorials, if you're satisfied with the overall structure/architecture of the update I could look into updating some of them, then we might get a better picture at how much work it actually ends up being.

est31:
It might be the best to work on the glutin upgrade on a separate branch, as it might be quite involved to port all the examples and the book. I have created a new `glutin-0.30` branch. Feel free to close this PR and file one against that one. Then we can merge it to master when the time is ready. I think if the utils module has less than 300-ish lines and is one file only then it's okay to include it as a separate file, but at least one or two examples should remain self contained.

Melchizedek6809:
Alright, I changed the base branch of this PR.
I'll look into adding an abstraction to the support crate and port an example to see if it fits in 300-ish lines.

What do you think about leaving the tutorial examples self contained?
Since I suppose most readers/users will end up building their own abstractions anyways.

est31:
> What do you think about leaving the tutorial examples self contained?

I'm okay with that but if it becomes too much copy-paste then it would be better to make them use a support module. This is good both from a reading point of view, you know you can skip those parts because they are the same in all examples and there are no slight example specific modifications, and also good from a maintenance point of view. The readers have top priority here.

Melchizedek6809:
> I'm okay with that but if it becomes too much copy-paste then it would be better to make them use a support module.

Alright I'll keep that in mind

> The readers have top priority here.

I respect that very much, I'll try my best.

Melchizedek6809:
@est31 Finally got around to looking at the examples, added a generic container that handles the event_loop and a trait to be implemented by the examples, the support module is currently at 173 lines, I hope that is alright.

Regarding the self-contained examples: there are a couple of examples that don't fit the usual schema (for example: gpgpu, info, screenshot, manual-creation, multiple_windows), do you think it's sufficient to have these be the self-contained ones?

examples/support/mod.rs
```diff
@@ -62,6 +62,7 @@ pub trait ApplicationContext {
     fn new(display: &Display<WindowSurface>) -> Self;
     fn update(&mut self) { }
     fn handle_window_event(&mut self, _event: &WindowEvent, _window: &winit::window::Window) { }
+    fn window_title() -> &'static str { "Glium example" }
```
est31, line 65: Can this be an associated constant instead? No need for a function. That also lets you avoid typing the `'static` every time.

Melchizedek6809: Sure, that makes sense. I've amended the last commit. Couldn't get rid of the lifetime specifier though.

est31: Thanks!

> Couldn't get rid of the lifetime specifier though.

Ahh yeah makes sense, there can be lifetimes in traits and such so it can't be omitted from there.

est31:
> Regarding the self-contained examples: there are a couple of examples that don't fit the usual schema (for example: gpgpu, info, screenshot, manual-creation, multiple_windows), do you think it's sufficient to have these be the self-contained ones?

Ideally, the triangle example would also be self-contained so that users can familiarize themselves with the entire structure. But maybe one should only do this for the book, so I guess it's fine to keep it here in the normal examples.

Melchizedek6809:
@est31 Alright, I've finally managed to port all the examples, most use the support crate except for tutorial1/2 as well as the SPIR-V example (it requires a higher OpenGL version than the rest). Those 3 also take some shortcuts to make things slightly less verbose.

I still have to check the book, I'm thinking about inserting a chapter in between 2 and 3 just dealing with context creation, since this probably requires an entire chapter to explain properly.

Apart from that I've fixed the tests so that they compile, they still fail as they do on the master branch, I suppose fixing them should probably go in another PR.

Also, I think we can remove the headless feature, since users can just initialize glutin without a window on Mac/Linux, so it doesn't seem to be required anymore, although I might be missing something, what do you think?

I'll also do some more testing on various machines to make sure that things work reliably since I've only really tested on Linux so far where things seem quite stable.

Melchizedek6809:
@est31 I've made a draft of the first chapter but have to say I'm quite dissatisfied with how complicated the initialization is (even though I've already simplified things), it also requires so many dependencies just to get things going. What do you think about introducing a convenience function that opens a window, creates a glutin context and returns the window/display pair? All behind a feature flag of course. That would make the first tutorial much more understandable in my opinion, since we can explain the details in a later chapter, and would look much more like `glium:master`.

As an example the code for the first tutorial currently looks like this:
```rust
use winit::event_loop::EventLoopBuilder;
use winit::window::WindowBuilder;
use glutin_winit::DisplayBuilder;
use glutin::config::ConfigTemplateBuilder;
use glutin::prelude::*;
use glutin::display::GetGlDisplay;
use glutin::surface::WindowSurface;
use raw_window_handle::HasRawWindowHandle;
use std::num::NonZeroU32;
use glium::Surface;

fn main() {
    // We start by creating the EventLoop, this can only be done once per process.
    // This also needs to happen on the main thread to make the program portable.
    let event_loop = EventLoopBuilder::new().build();

    // First we start by opening a new Window
    let window_builder = WindowBuilder::new().with_title("Glium tutorial #1");
    let display_builder = DisplayBuilder::new().with_window_builder(Some(window_builder));
    let config_template_builder = ConfigTemplateBuilder::new();
    let (window, gl_config) = display_builder
        .build(&event_loop, config_template_builder, |mut configs| {
            // Just use the first configuration since we don't have any special preferences right now
            configs.next().unwrap()
        })
        .unwrap();
    let window = window.unwrap();

    // Now we get the window size to use as the initial size of the Surface
    let (width, height): (u32, u32) = window.inner_size().into();
    let attrs = glutin::surface::SurfaceAttributesBuilder::<WindowSurface>::new().build(
        window.raw_window_handle(),
        NonZeroU32::new(width).unwrap(),
        NonZeroU32::new(height).unwrap(),
    );

    // Finally we can create a Surface, use it to make a PossiblyCurrentContext and create the glium Display
    let surface = unsafe { gl_config.display().create_window_surface(&gl_config, &attrs).unwrap() };
    let context_attributes = glutin::context::ContextAttributesBuilder::new().build(Some(window.raw_window_handle()));
    let current_context = Some(unsafe {
        gl_config.display().create_context(&gl_config, &context_attributes).expect("failed to create context")
    }).unwrap().make_current(&surface).unwrap();
    let display = glium::Display::from_context_surface(current_context, surface).unwrap();

    // Start rendering by creating a new frame
    let mut frame = display.draw();
    // Which we fill with an opaque blue color
    frame.clear_color(0.0, 0.0, 1.0, 1.0);
    // By finishing the frame swap buffers and thereby make it visible on the window
    frame.finish().unwrap();

    // Now we wait until the program is closed
    event_loop.run(move |event, _, control_flow| {
        match event {
            winit::event::Event::WindowEvent { event, .. } => match event {
                // This event is sent by the OS when you close the Window, or request the program to quit via the taskbar.
                winit::event::WindowEvent::CloseRequested => control_flow.set_exit(),
                _ => (),
            },
            _ => (),
        };
    });
}
```

I'd like to change this to the following:

```rust
use winit::event_loop::EventLoopBuilder;
use glium::Surface;

fn main() {
    // We start by creating the EventLoop, this can only be done once per process.
    // This also needs to happen on the main thread to make the program portable.
    let event_loop = EventLoopBuilder::new().build();
    let (window, display) = glium::Display::simple_winit_display(&event_loop, "Glium tutorial #1");

    // Start rendering by creating a new frame
    let mut frame = display.draw();
    // Which we fill with an opaque blue color
    frame.clear_color(0.0, 0.0, 1.0, 1.0);
    // By finishing the frame swap buffers and thereby make it visible on the window
    frame.finish().unwrap();

    // Now we wait until the program is closed
    event_loop.run(move |event, _, control_flow| {
        match event {
            winit::event::Event::WindowEvent { event, .. } => match event {
                // This event is sent by the OS when you close the Window, or request the program to quit via the taskbar.
                winit::event::WindowEvent::CloseRequested => control_flow.set_exit(),
                _ => (),
            },
            _ => (),
        };
    });
}
```

This would also mean that users would only have to directly depend on glium and winit, which should be easy enough to explain, also users wouldn't have to write unsafe code.

kchibisov:
Be aware that your context is separate from the surfaces, you don't even need a surface to draw in some cases.

You can surely add some wrapper, but be aware that low level GL libs are mostly used by game engines, toolkits, and so on, so they hide all the complexity anyway (or they already have pretty complex init).

Optimizing for the examples is not really a good thing, though, sharing code between examples is reasonable. It would be just weird if glium will have direct dependency on winit, which it shouldn't, as glutin doesn't depend on winit at all.

Melchizedek6809:
Yeah in this PR glium already doesn't directly depend on winit anymore (apart from the examples of course), this would mostly serve to simplify the tutorials, since advanced users as you said probably have special requirements and do their own initialization anyways. I just fear that beginners might get discouraged by all the initialization they need to write to get a window/context going, since they don't have any need for that in the beginning.

est31:
> What do you think about introducing a convenience function that opens a window, creates a glutin context and returns the window/display pair? All behind a feature flag of course.

Yes, that would be wonderful. The second proposed snippet looks way better. Big :+1: from me. Also it would be good to [enable the feature on docs.rs](https://docs.rs/about/metadata) if it's not on by default.

Melchizedek6809:
> Yes, that would be wonderful. The second proposed snippet looks way better. Big +1 from me. Also it would be good to [enable the feature on docs.rs](https://docs.rs/about/metadata) if it's not on by default.

Great to heart that, I'll work on this next then

> (Edit: also, ideally, it would be really nice if you could set simple things like the window size and call other stuff on the winit WindowBuilder, with also as little boilerplate as possible)

Alright, sounds good, I'll do that

Melchizedek6809:
@est31 
Alright, I've gone over the entire book and updated it all to be compatible with glutin 0.30, apart from that I've made some changes to the tutorials:

- Instead of enforcing 60Hz the tutorials now render in response to a `RedrawRequested` event (and request a redraw on `RedrawEventsCleared`)
- The animated triangles are now positioned using a sine function, this makes the animation much more apparent, since before it was pretty hard to see on a high resolution monitor.
- In chapter 5 the color is now being set as a vertex attribute instead of using `position.xy` as rg values, this way the result looks nicer to me and seemed more useful for readers to know.
- In chapter 6 where textures are introduced I've changed the shape to a square, that way the OpenGL logo isn't cut off/distorted which should make for a nicer end result.

I've also fixed some typos in various places and reworded some sentences so there's less room for misunderstandings.

The update in general also seems quite stable since I haven't encountered any problems in the past weeks on various machines that I have access to, would be nice if you could do a final review of the changes.

Melchizedek6809:
@est31 Hey there, just wanted to ask whether you've found anything that needs changing yet :)

est31:
> would be nice if you could do a final review of the changes.

Okay thanks I somehow missed that this was ready for review. I am a bit busy with work and other projects atm but I will give it a review in the coming days, hopefully.

src/lib.rs
```diff
@@ -14,11 +14,11 @@ The initialisation of a glium display occurs in several steps.
 extern crate glium;
 
 fn main() {
-    // 1. The **winit::EventsLoop** for handling events.
-    let mut events_loop = glium::glutin::event_loop::EventLoop::new();
+    // 1. The **winit::EventLoop** for handling events.
+    let mut events_loop = glium::winit::event_loop::EventLoop::new();
     // 2. Parameters for building the Window.
-    let wb = glium::glutin::window::WindowBuilder::new()
-        .with_inner_size(glium::glutin::dpi::LogicalSize::new(1024.0, 768.0))
```
est31, line 21: Is this example still accurate?

src/vertex_array_object.rs
```diff
@@ -1,4 +1,4 @@
-use std::borrow::Borrow;
+use std::borrow::{Borrow};
```
est31, line 1: Can you un-do this change?

est31 (changes requested):
Sorry for the late review, it looks really good already! Thanks for putting in all the work.

src/backend/glutin/mod.rs
```diff
@@ -287,9 +281,50 @@ unsafe impl Backend for GlutinBackend {
...
     }
 }
+
+#[cfg(feature = "simple_winit_window")]
+/// Create a new winit Window and create a glium Display for it.
+///
+/// This function is meant as an easy way to get started> If you require more
+/// flexibility, you need to initialize the window and context yourself.
+/// Have a look at the source of this function as an example on how to do that.
+pub fn simple_winit_window(event_loop: &winit::event_loop::EventLoop<()>, window_title: &str) -> (winit::window::Window, Display<glutin::surface::WindowSurface>) {
```
est31, line 295: I think this function is great, but ideally there would additionally be a way to adjust the window size. E.g. by letting people pass in the window builder. I think it would be best to have that either in a separate function, or via the builder pattern, i.e. `SimpleWindowBuilder::new()` and `SimpleWindowBuilder::window_builder(...)`, and `SimpleWindowBuilder::build()` etc. This can be something for the future though, after this PR is merged.

src/backend/glutin/mod.rs
```diff
@@ -287,9 +281,50 @@ unsafe impl Backend for GlutinBackend {
...
-        Takeable::insert(&mut gl_window_takeable, new_gl_window);
+        let pair = self.borrow();
+        pair.context.make_current(&pair.surface).unwrap();
     }
 }
+
+#[cfg(feature = "simple_winit_window")]
+/// Create a new winit Window and create a glium Display for it.
+///
+/// This function is meant as an easy way to get started> If you require more
```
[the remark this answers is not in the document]
Melchizedek6809: Alright, should be fixed now

book/tuto-01-getting-started.md
```diff
@@ -1,90 +1,73 @@
...
-    let wb = glutin::window::WindowBuilder::new();
-    let cb = glutin::ContextBuilder::new();
-    let display = glium::Display::new(wb, cb, &event_loop).unwrap();
+    let event_loop = winit::event_loop::EventLoopBuilder::new().build();
+    let (_window, display) = glium::backend::glutin::simple_winit_window(&event_loop, "Glium tutorial #1");
 }
 ```
 
-But there is a problem: as soon as the window has been created, our main function exits and `display`'s destructor closes the window. To prevent this, we need to loop forever until we detect that a `CloseRequested` event has been received. We do so by calling `event_loop.run`:
+If you try to run this example with `cargo run` you'll encounter  a problem: as soon as the window has been created, our main function exits and the window is closed. To prevent this, we need to wait until we receive a `CloseRequested` event. We do this by calling `event_loop.run`:
```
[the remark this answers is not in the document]
Melchizedek6809: Alright, should be fixed now

book/tuto-02-triangle.md
```diff
@@ -24,15 +24,20 @@ Each triangle is made of three vertices, which means that a shape is just a coll
 struct Vertex {
     position: [f32; 2],
 }
-
 implement_vertex!(Vertex, position);
 ```
 
-Our struct contains a `position` field which we will use to store the position of each vertex on the window. Being a true vectorial renderer, OpenGL doesn't use coordinates in pixels. Instead it considers that the window has a width and a height of 2 units, and that the origin is at the center of the window.
+To make the implement_vertex! macro available you need to add the following lines above the main function:
```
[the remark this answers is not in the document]
Melchizedek6809: Alright, should be fixed now

src/backend/glutin/mod.rs
```diff
@@ -9,161 +9,175 @@ Only available if the 'glutin' feature is enabled.
...
 use std::fmt;
+use std::num::NonZeroU32;
 use std::ops::Deref;
 use std::os::raw::c_void;
 use std::rc::Rc;
-use takeable_option::Takeable;
-use crate::{Frame, IncompatibleOpenGl, SwapBuffersError};
+use crate::{Frame, IncompatibleOpenGl};
+
+/// Wraps glutin context with a the Surface
```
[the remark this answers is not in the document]
Melchizedek6809: Alright, should be fixed now

src/lib.rs
Melchizedek6809: Not really, I've rewritten it using `simple_winit_window`, I hope that's alright, since otherwise it'd get pretty long.

src/vertex_array_object.rs
Melchizedek6809: Sure

src/backend/glutin/mod.rs
Melchizedek6809: I went ahead and added a `SimpleWindowBuilder` with 2 methods for specifying title and size, as well as one for replacing the winit `WindowBuilder`. Because of this `simple_winit_window` isn't really necessary anymore and I've removed it and changed the tutorials to reflect that. Thought it might be good to have a builder from the start so that we can extend it without breaking existing code.

Melchizedek6809:
@dcvz No problem and thanks for the offer and reminding me, I might take you up on that if there's much to do in the next round.

src/lib.rs
est31: Can you add back the function calls here on the `SimpleWindowBuilder`?

est31 (approved):
Thanks a lot for this monumental effort. This has been one of the biggest changes in `glium` in the last 5 years I think.

PR #2041, opened 2023-01-21
Support GL_ARB_clip_control

timlathy:
>  [This extension provides additional clip control modes to configure how clip space is mapped to window space.  This extension's goal is to 1) allow OpenGL to effectively match Direct3D's coordinate system conventions, and 2) potentially improve the numerical precision of the Z coordinate mapping.](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_clip_control.txt)


est31 (approved):
Thanks!

Issue #2044, opened 2023-03-09
DepthRenderBuffer errors with zero size


justincredible:
[error_renderbuffer.txt](https://github.com/glium/glium/files/10928723/error_renderbuffer.txt)

Windows 10
NVIDIA GeForce GTX 1660
glium 0.32.1

This error output appears on use of a `DepthRenderBuffer` created with either a zero width or height. The attached file was generated from the FXAA example by minimizing the window, but resizing the window height to zero has the same result.

It is easy enough to avoid using or even creating a zero sized `DepthRenderBuffer`, but I was surprised that I do not see errors when the `Texture2d` is created with a zero width or height. When calling OpenGL directly, creating a framebuffer with a zero-sized texture or renderbuffer results in a framebuffer status of GL_FRAMEBUFFER_INCOMPLETE_ATTACHMENT.

Melchizedek6809:
Thanks for reporting this issue, took me a while to reproduce it since it only seems to occur on Windows. The linked PR should fix that issue.

justincredible:
Yes, the PR works well. Thanks for addressing this.

Melchizedek6809:
Thank you for testing!
Seems to be the same issue, I've added a fix to the existing PR, the compute_image example also seems to have issues when minimized on Windows, though it seems to have a different cause.

justincredible:
That's good to know, I've run the compute shader examples but have not really looked into them yet.

PR #2051, opened 2023-04-01
Validate that a framebuffer isn't 0 on any dimension

Melchizedek6809:
I've added a check during framebuffer creation that makes sure that it's not 0 on any dimension, since this would lead to other errors later, now we return a ValidationError::EmptyFramebufferUnsupportedDimensions in that case.

Since on Windows one can resize a window to have 0 width/height which also happens when minimizing I've changed the fxaa example so that it doesn't crash on Windows (Mac and Linux with various WMs seem to ensure that a window is alwasy at least 1x1)

This fixes https://github.com/glium/glium/issues/2044

### DIFF
--- a/examples/fxaa.rs
+++ b/examples/fxaa.rs
@@ -157,7 +157,15 @@ mod fxaa {
     pub fn draw<T, F, R>(system: &FxaaSystem, target: &mut T, enabled: bool, mut draw: F)
                          -> R where T: Surface, F: FnMut(&mut SimpleFrameBuffer<'_>) -> R
     {
-        let target_dimensions = target.get_dimensions();
+        let mut target_dimensions = target.get_dimensions();
+        // We need to ensure that our framebuffer is at least 1x1. Otherwise, we would get an error.
+        // This could happen when minimizing or resizing the window on Windows, for example.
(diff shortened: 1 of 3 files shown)


examples/fxaa.rs
```diff
@@ -157,7 +157,15 @@ mod fxaa {
     pub fn draw<T, F, R>(system: &FxaaSystem, target: &mut T, enabled: bool, mut draw: F)
                          -> R where T: Surface, F: FnMut(&mut SimpleFrameBuffer<'_>) -> R
     {
-        let target_dimensions = target.get_dimensions();
+        let mut target_dimensions = target.get_dimensions();
+        // We need to ensure that our framebuffer is at least 1x1, otherwise we would get an error, this can
+        // happen when you minimize/resuze the window on Windows for example.
```
est31, line 162: typo

est31 (changes requested):

examples/fxaa.rs
Melchizedek6809: Ah my bad, I amended the commit.

est31 (approved):
Thanks!

Issue #1440, opened 2016-03-22
array_layout_offsets test fails on x86_64-pc-windows-gnu


brson:
I see this output when running cargo test:

```
failures:

---- array_layout_offsets stdout ----
        thread 'array_layout_offsets' panicked at 'assertion failed: `(left == right)` (left: `Struct { members: [("data", Array { content: Struct { members: [("pos", BasicType { ty: FloatVec2, offset_in_buffer: 928 }), ("speed", BasicType { ty: Float, offset_in_buffer: 2896 }), ("dir", BasicType { ty: FloatVec2, offset_in_buffer: 1256 })] }, length: 256 })] }`, right: `Struct { members: [("data", Array { content: Struct { members: [("pos", BasicType { ty: FloatVec2, offset_in_buffer: 0 }), ("dir", BasicType { ty: FloatVec2, offset_in_buffer: 8 }), ("speed", BasicType { ty: Float, offset_in_buffer: 16 })] }, length: 256 })] }`)', tests\shaders.rs:831
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    array_layout_offsets
```

My rustc is 

```
$ rustc -Vv
rustc 1.9.0-nightly (21922e1f4 2016-03-21)
binary: rustc
commit-hash: 21922e1f48b263b39498cfeec79c1ca3f5886efe
commit-date: 2016-03-21
host: x86_64-pc-windows-gnu
release: 1.9.0-nightly
```

Glium commit is 6aadf08fe39547ea99bac02c2943b2975498eb33

tomaka:
Oh it's the third time that this test fails already.
Offsets are tricky to get right, and I'm never sure whether it's a problem in glium, in glium's test, or in the driver.

tomaka:
Pretty printed:

> Struct { members: [("data", Array { content: Struct { members: [("pos", BasicType { ty: FloatVec2, offset_in_buffer: 928 }), ("speed", BasicType { ty: Float, offset_in_buffer: 2896 }), ("dir", BasicType { ty: FloatVec2, offset_in_buffer: 1256 })] }, length: 256 })] }

> Struct { members: [("data", Array { content: Struct { members: [("pos", BasicType { ty: FloatVec2, offset_in_buffer: 0 }), ("dir", BasicType { ty: FloatVec2, offset_in_buffer: 8 }), ("speed", BasicType { ty: Float, offset_in_buffer: 16 })] }, length: 256 })] }`)

[...]

PR #2052, opened 2023-04-22
Tests on Windows

justincredible:
Hi,

I am aware that the tests not working on Windows has been a [known issue](https://github.com/glium/glium/issues/1780) for [several years](https://github.com/glium/glium/issues/1440) now. The advice to use `--test-threads 1` is no longer applicable as it appears to spin up new threads rather than forcing them all to run on the main thread. The panic message from `EventLoop` suggests using the extension method `with_any_thread` on `EventLoopBuilder`, which does allow running a single test per process via `cargo test --test <Target> <Name> -- --exact`.

The changes to the test modules fix the failures on my system. The change to support/mod.rs is of questionable value outside a Windows platform. There are only six remaining tests that are still failing:
- `multiple_displays` and `bindless_texture_residency_context_rebuild` can both be made to pass by using a singleton event_loop. I do not think this can be done in Safe Rust, and the singleton does not help with running multiple tests per process anyway
- I am not sure the four `empty_stenciltexture(1|2)d(array)*` tests are meant to be uncommented. I did not attempt to get them passing
- I am finding 322 active test names in the tests folder, but I may have missed something

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -77,14 +77,15 @@ macro_rules! uniform {
 /// ## Example
 ///
 /// ```rust
+/// # #[macro_use] extern crate glium;
 /// # use glium::uniform;
 /// # fn main(){
-///     let uniforms = dynamic_uniform!{
-///         color: [1.0, 1.0, 0.0, 1.0],
-///         some_value: 12i32,
+///     let mut uniforms = dynamic_uniform!{
+///         color: &[1.0, 1.0, 0.0, 1.0],
+///         some_value: &12i32,
 ///     };
 ///
-///     uniforms.add("another_value", 1.5f32);
+///     uniforms.add("another_value", &1.5f32);
 /// # }
 /// ```
 ///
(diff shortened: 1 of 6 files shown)


est31:
Thanks for filing a PR! It would be great if we could run the tests in CI.

FYI a large refactor of glium's tests/examples is going on in #2036 . It would be best to coordinate with that PR.

justincredible:
Thanks, I'll coordinate with that PR while it's being reviewed.

justincredible:
This is still a WIP but I have a POC for running tests in Windows via `cargo test`.

Please do *not* approve the workflows as this still has a non-portable implementation and will fail on Ubuntu.

I am not sure this approach can work with `rebuild_display`, but this only affects one test. I will also need to look at the four failing `empty_stenciltexture*` tests, but otherwise `cargo test --tests` runs successfully.

Now that I can run `cargo test`, I am seeing some doc test failures. There are also some build warnings for the tests. Once these are addressed, this approach should work easily with PR 2036 and I can merge there if it's still under review.

justincredible:
Closing this again as I have a PR against #2036.

PR #2054, opened 2023-04-23
Fix mdbook hiding `#version..` in rust codeblocks

### LINKED ISSUE #2053: `#version 140` in book codeblocks is hidden by rustdoc
Easy fix if the intention is for examples to only be viewable using mdbook.

Could not find contribution guidelines, but am willing to just put a PR as its a really small fix/issue.
[see code block in this section](https://glium.github.io/glium/book/tuto-02-triangle.html#program).

The `#` needs to be escaped so that it is `## version 140`, otherwise Rustdoc / mdbook hides lines starting with `#`. See Rustdoc docs [here](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example)

Could be a source of confusion.

ChristopherPerry6060:
Fixes #2053. This does not touch the glsl code blocks as they seem to be left alone.

Please note there might be a more inclusive fix for this but, this will fix it for anyone viewing through the link on the repo. 

If viewing the raw markdown files it will show as `##version`.

### DIFF
--- a/book/tuto-02-triangle.md
+++ b/book/tuto-02-triangle.md
@@ -71,7 +71,7 @@ The tricky part is that *we* need to write the vertex and fragment shaders. To d
 
 ```rust
 let vertex_shader_src = r#"
-    #version 140
+    ##version 140
 
     in vec2 position;
 
@@ -91,7 +91,7 @@ The second shader is called the fragment shader (sometimes also named *pixel sha
 
 ```rust
 let fragment_shader_src = r#"
-    #version 140
+    ##version 140
 
     out vec4 color;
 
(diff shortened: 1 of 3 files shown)


Melchizedek6809:
I'm wondering couldn't we use the `program!` macro in the book to work around this issue? That way the markdown source could be pasted as is.

est31 (approved):
Thanks!

est31:
We don't really have GLSL syntax highlighting for these snippets as they are inside rust raw strings.

I think replacing those strings with the `program!` macro is a nice idea, but IMO that's something for a different/later PR.

est31:
(also, FTR, this is for the book, which is generated by `mdbook` and not `rustdoc` but that's only a detail)

PR #2055, opened 2023-04-28
PS Vita support

nikarh:
Hi, I am not sure that platform-specific additions are in the course of this project, but anyway,

Recently I ported Rust std to PS Vita via homebrew newlib (these PR's are still being reviewed and probably will be merged soon).
In PS Vita homebrew world there exists a GLES 2 implementation which is essentially a facade above the platforms graphics API.

In this PR I tried to make `glium` work on PS Vita homebrew with this implementation.

What exactly is changed?
- `TextureFormat::get_formats_list()` - ~I am not entirely sure what would be the best approach with this one. The iterator changing just crashes on PS Vita (and I wasn't able to fix that by increasing the stack size). Obviously allocating multiple Vec's on the heap to collect them all is not the best idea, and this function honestly should return some static slice instead of a Vec, but considering that this Vec is just constructed twice when glium is initialized, this might not be that huge of a deal. I looked into some macros that may help with this one, like strum's [EnumIter](https://docs.rs/strum_macros/0.24.3/strum_macros/derive.EnumIter.html), but unfortunately it doesn't flatmap inner enum values.~
- Internally on PS Vita all vertex attributes are assumed as [float;4], so I had to disable the attribute type check for Vita
- Uniform location check used to check if it is >=0. The [spec](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glGetUniformLocation.xhtml) technically only defines `-1` as an error value. Vita GLES library uses negative values there.
- ~And there's a bug in Vita GLES wrapper, which sets only one integer on MAX_VIEWPORT_DIMS, yielding (4096;0), instead of a square~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ backtrace = "0.3.2"
 lazy_static = "1.0"
 smallvec = "1.0"
 fnv = "1.0.5"
+cfg-if = "1.0.0"
 
 [build-dependencies]
 gl_generator = "0.14"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,6 @@ backtrace = "0.3.2"
 lazy_static = "1.0"
 smallvec = "1.0"
 fnv = "1.0.5"
-cfg-if = "1.0.0"
 
 [build-dependencies]
 gl_generator = "0.14"
--- a/src/image_format.rs
+++ b/src/image_format.rs
@@ -1418,17 +1418,17 @@ impl TextureFormat {
     /// Returns a list of all the possible values of this enumeration.
     #[inline]
     pub fn get_formats_list() -> Vec<TextureFormat> {
-        // TODO: this function looks ugly
-        UncompressedFloatFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        UncompressedIntFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        UncompressedUintFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        SrgbFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        CompressedFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        CompressedSrgbFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        DepthFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        StencilFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        DepthStencilFormat::get_formats_list().into_iter().map(|f| f.to_texture_format())))))))))
-        .collect()
+        // Chaining so many iterators can blow up the stack on some platforms
+        let mut result: Vec<_> = UncompressedFloatFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).collect();
+        result.append(&mut UncompressedIntFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).collect());
+        result.append(&mut UncompressedUintFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).collect());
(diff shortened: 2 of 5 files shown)


est31:
Thanks for the PR! Generally, supporting additional platforms is welcome, as long as it's not too complicated to do so, requires NDAs or such.

> there's a bug in Vita GLES wrapper, which sets only one integer on MAX_VIEWPORT_DIMS, yielding (4096;0), instead of a square

Instead of doing a workaround here, can you file a bug upstream? I suppose you are using [this](https://github.com/Rinnegatamante/vitaGL)?

src/image_format.rs
```diff
@@ -1418,17 +1418,17 @@ impl TextureFormat {
...
-        SrgbFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        CompressedFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        CompressedSrgbFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        DepthFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        StencilFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).chain(
-        DepthStencilFormat::get_formats_list().into_iter().map(|f| f.to_texture_format())))))))))
-        .collect()
+        // Chaining so many iterators can blow up the stack on some platforms
+        let mut result: Vec<_> = UncompressedFloatFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).collect();
+        result.append(&mut UncompressedIntFormat::get_formats_list().into_iter().map(|f| f.to_texture_format()).collect());
```
est31, line 1423: Can you use extend here? https://doc.rust-lang.org/stable/std/iter/trait.Extend.html#tymethod.extend

est31 (changes requested):

src/image_format.rs
nikarh: Absolutely, will do

nikarh:
I'll try to get the fix with GL_MAX_VIEWPORT_DIMS to upstream and get back here when it gets merged or denied

nikarh:
Oh, I can always use a fork of `vitaGL`, so I'll remove this specific hack from this PR regardless

nikarh:
For the sake of history, here's the PR to `vitaGL`

Rinnegatamante/vitaGL#102

src/vertex_array_object.rs
```diff
@@ -277,11 +277,17 @@ impl VertexArrayObject {
...
 
-                if ty.get_num_components() != attribute.ty.get_num_components() ||
-                    attribute.size != 1
-                {
-                    panic!("The program attribute `{}` does not match the vertex format. \
-                            Program expected {:?}, got {:?}.", name, attribute.ty, ty);
+                cfg_if::cfg_if! {
+                    // Unfortunately internal API used by GLES implementation on Vita
+                    // assumes all attributes as float4, so we should skip this check for it.
+                    if #[cfg(not(target_os = "vita"))] {
```
est31, line 283: Can you do `if cfg!` instead of this?

src/uniforms/bind.rs
```diff
@@ -296,7 +296,10 @@ fn bind_uniform<P>(ctxt: &mut context::CommandContext,
                    name: &str)
                    -> Result<(), DrawError> where P: ProgramExt
 {
-    assert!(location >= 0);
+    // The spec only states that a negative location is an error.
+    // This fact is abused by some implementations (e.g. vitaGL)
+    // which actually use negative values for uniform locations.
+    assert!(location != -1);
```
est31, line 302: Same here, ideally one would only make this check less strict on non-vitaGL platforms.

Cargo.toml
```diff
@@ -44,6 +44,7 @@ backtrace = "0.3.2"
 lazy_static = "1.0"
 smallvec = "1.0"
 fnv = "1.0.5"
+cfg-if = "1.0.0"
```
est31, line 47: This dependency isn't needed.

est31 (changes requested):
Looks very good already :+1: except of the things I've pointed out

est31 (approved):
Thanks!

PR #2056, opened 2023-05-03
Reduce memory allocations in draw

nikarh:
Hi,
This PR reduces memory allocations in draw calls by doing two things:

1. By changing `VertexFormat` from `Cow<'static, [...]>` to a `&'static [...]`.
2. Changing `vaos` from Vec to SmallVec of 3 elements.

## VertexFormat
Now, as far as I can tell from the comments in the code, this is something that the original author wanted to.

What's wrong with the `Cow`? Well, first of all, it's useless - Cow::Borrowed was never constructed, and in every draw call it was [cloned](https://github.com/glium/glium/blob/9045bc1e9e4bbcc149be29078f28921ac413cd28/src/vertex_array_object.rs#LL152C64-L152C64), and `Cow::Owned::clone` clones the data, meaning it cloned the `Vec`.

How it's possible to use a static slice as a vertex format?
The main problem was adapting `implement_vertex` macro. Essentially a macro now has to create a const or static field, and in the trait implementation return it. To do this, the whole logic behind this const generation should also be const. This implies two things:
1. Calculating field offset (`__glium_offset_of`). This is easily doable now, since `offset_from` on pointers is const in rust stable. The latest `memoffset` crate uses it, so `__glium_offset_of` is now also const
2. Making `attr_type_of_val` const. This was a bit more complicated because it called a trait fn, and trait fn's can't be const. So I've adapted the trait, and now it has a const field. The other thing is `Option::map` which is not const, so I had to rewrite that to match.

## vaos

The only place that writes vaos is [this](https://github.com/glium/glium/blob/9045bc1e9e4bbcc149be29078f28921ac413cd28/src/vertex_array_object.rs#L188), and `self.vertex_buffers` is already a `SmallVec` with 2 elements. Only one element is pushed there two lines later, meaning this Vec will always have a constant 3-element size. By changing data type we reduce 2 allocations (one in collect, and one in vec resize in push)

## glium_derive

glium/glium_derive#7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ features = []
 optional = true
 
 [dependencies]
-memoffset = "0.6"
+memoffset = "0.8"
 takeable-option = "0.5"
 backtrace = "0.3.2"
 lazy_static = "1.0"
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -156,104 +156,106 @@ macro_rules! dynamic_uniform{
 #[macro_export]
 macro_rules! implement_vertex {
     ($struct_name:ident, $($field_name:ident),+) => (
+        impl $struct_name {
+            const BINDINGS: $crate::vertex::VertexFormat = &[
+                $(
+                    (
+                        std::borrow::Cow::Borrowed(stringify!($field_name)),
+                        $crate::__glium_offset_of!($struct_name, $field_name),
+                        -1,
+                        {
+                            const fn attr_type_of_val<T: $crate::vertex::Attribute>(_: Option<&T>)
+                                -> $crate::vertex::AttributeType
+                            {
+                                <T as $crate::vertex::Attribute>::TYPE
+                            }
+                            let field_option = match None::<&$struct_name> {
+                                Some(v) => Some(&v.$field_name),
+                                None => None
+                            };
+                            attr_type_of_val(field_option)
+                        },
+                        false
+                    )
+                ),+
+            ];
+        }
+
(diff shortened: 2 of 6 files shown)


nikarh:
By the way, the only apparent allocation that is remaining in `draw` call is the allocation and resizing of fences Vector.
Since this vec contains a structure with a lifetime, I don't really see a way of reusing the vector without using some crazy unsafe magic:
 - E.g. keeping a cached Vec with `'static` lifetime, and then doing `mem::transmute` to effectively cast it to a smaller lifetime
 - And then since `fence.insert` requires `self` to be moved, if we want to iterate the vec in the same order linearly, in the loop which consumes fences, I would have to resort to some `mem::swap` to `MaybeUninit::zeroed`. Or change the signature to Option (probably better approach). Or change `insert` signature to accept mut ref (probably a bad idea, because it would make multiple inserts possible). Or change the API, extract fences as some struct with methods, and hide Vec of options behind the implementation (probably the best approach).

Or just ignore it, because it would be one allocation per draw call (if I'm lucky and there are just two fences (as far as I understand vec.push grows capacity exponentially).

Though I believe 0 allocations on draw call is still a good thing to have. Especially for some very old and odd hardware.

Melchizedek6809:
This looks great! Thank you very much for this.

Regarding the fence allocations: hiding the underlying datastructure for fences in a struct sounds like a good approach, do you want to give it a try?

And 0 allocations on draw calls would be fantastic, not just for old/odd hardware but also underpowered ARM SBC's

PR #2061, opened 2023-06-24
Fix new_raw vertex buffer creation

### LINKED ISSUE #1461: VertexBuffer::new_raw() example doesn't seem to work
I tried to draw some lines (as a LineStrip) from a float array of consecutive x and y coordinates following [the example for VertexBuffer::new_raw](https://tomaka.github.io/glium/glium/vertex/struct.VertexBuffer.html#method.new_raw), but the 3 vertices I've given (red lines) were interpreted as 6 vertices (cyan lines).

```rust
let bindings = Cow::Owned(vec![(
        Cow::Borrowed("position"), 0,
        glium::vertex::AttributeType::F32F32,
    ),
]);
let data = [
    -0.5_f32, -0.5_f32,
     0.0,  0.5,
     0.5, -0.25,
];
let vertex_buffer = (unsafe {
    glium::VertexBuffer::new_raw(
        &display, &data, bindings, 2 * ::std::mem::size_of::<f32>())
}).unwrap();
```

However, when I changed `data` to an array of arrays (or tuples) it worked as expected:

```rust
let data = [
    [-0.5_f32, -0.5_f32],
    [ 0.0,  0.5],
    [ 0.5, -0.25],
 ];
```

Also, it seems that [the last argument of VertexBuffer::new_raw](https://github.com/tomaka/glium/blob/5a88c24993c20cddc4ca719edaee6d29584efd41/src/vertex/buffer.rs#L255), `elements_size`, is not used at all.

screenshot - 250416 - 13 39 12

dcvz:
Fixes #1461 - grabbed from a linked commit in that issue written by @mjk3979 

This was causing an issue as well for me as the size was not being taken into consideration

### DIFF
--- a/src/buffer/view.rs
+++ b/src/buffer/view.rs
@@ -1065,6 +1065,24 @@ pub struct BufferAny {
 }
 
 impl BufferAny {
+    /// Builds a new buffer containing the given data. The size of the buffer is equal to the size
+    /// of the data.
+    pub fn new<F: ?Sized, T: Content + ?Sized>(facade: &F, data: &T, ty: BufferType, mode: BufferMode, elements_size: usize)
+                  -> Result<BufferAny, BufferCreationError>
+                  where F: Facade
+    {
+        Alloc::new(facade, data, ty, mode)
+            .map(|buffer| {
+                let size = buffer.get_size();
+                BufferAny {
+                    alloc: buffer,
+                    size,
+                    elements_size,
+                    fence: Fences::new(),
+                }
+            })
+    }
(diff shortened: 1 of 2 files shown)


est31:
Outside of the normal review, if you copied their code, and @mjk3979 has not submitted it to glium in a PR, we need their consent to merge this. @mjk3979 do you agree with this being merged?

est31:
Also on top of that, authorship information was replaced. Ideally the commit would have the original author, @mjk3979.

PR #2067, opened 2023-07-15
Apply fix to resurrected tests

justincredible:
I wasn't able to track down how these two tests disappeared in the glutin 0.30 branch, but a rebase several weeks before it was merged is what resurrected them.

This change applies to these two tests a previously merged fix 4add826bcb34cfb6f5fb1ed38215cbb8ddc4101f for the other two tests that use the (1.0, 0.0, 0.0, 0.5) colour value.


est31 (approved):
Thanks! And sorry for the inconvenience of having this test broken again.

PR #2068, opened 2023-07-19
Running tests on Unix and Windows

justincredible:
Attempting to run `cargo test --tests` or `cargo test [--all] --all-targets` quickly results in failure with:

> thread 'basic' panicked at 'Initializing the event loop outside of the main thread is a significant cross-platform compatibility hazard. If you absolutely need to create an EventLoop on a different thread, you can use the EventLoopBuilderExtUnix::any_thread function.', /home/justin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.28.6/src/platform_impl/linux/mod.rs:697:13

Following the error message's advice leads to the next error:

> thread 'attribute_float_i16' panicked at 'Creating EventLoop multiple times is not supported.', /home/justin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.28.6/src/event_loop.rs:116:13

This change addresses both issues on Unix and Windows by creating the event loop on its own thread and setting up communication channels so new windows can be sent to test threads on request. On Wayland, the `WindowBuilder` option `with_visible(false)` is not respected, but this is just an annoyance and does not cause issues running the tests. On X11, there are additional issues that prevent the tests from simply being run:
- All tests under `tests/attrs.rs` hang indefinitely.
- The other test targets can run to completion but are often met with the intermittent failure `error: XError(XError { description: "GLXBadWindow", error_code: 170, request_code: 152, minor_code: 32 })` that causes any remaining tests in the target to also fail

The implementation has a bit of a code smell, but it allows the changes to be localised around the `build_display` function and the call sites do not need to be updated. I have no experience with GitHub Actions, but I attempted to update the CI and was met with the errors:

Ubuntu
> thread 'event_loop' panicked at 'Failed to initialize any backend! Wayland status: NoCompositorListening X11 status: XOpenDisplayFailed', /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.28.6/src/platform_impl/linux/mod.rs:757:9

Windows
> thread 'basic' panicked at 'called `Result::unwrap()` on an `Err` value: IncompatibleOpenGl("OpenGL implementation doesn't support buffer objects\nOpenGL implementation doesn't support vertex/fragment shaders\nOpenGL implementation doesn't support framebuffers\nOpenGL implementation doesn't support blitting framebuffers")', tests\support\mod.rs:139:75

### DIFF
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -20,12 +20,22 @@ use glutin::surface::{SurfaceAttributesBuilder, WindowSurface};
 use glutin_winit::DisplayBuilder;
 use raw_window_handle::HasRawWindowHandle;
 
+// There is a Wayland version of this extension trait but the X11 version also works on Wayland
+#[cfg(unix)]
+use winit::platform::x11::EventLoopBuilderExtX11;
+#[cfg(windows)]
+use winit::platform::windows::EventLoopBuilderExtWindows;
+
 use std::env;
 
 /// Builds a display for tests.
 pub fn build_display() -> Display<WindowSurface> {
     let version = parse_version();
-    let event_loop = EventLoopBuilder::new().build();
+    let event_loop = if cfg!(unix) || cfg!(windows) {
+        EventLoopBuilder::new().with_any_thread(true).build()
+    } else {
+        EventLoopBuilder::new().build()
+    };
     let window_builder = WindowBuilder::new().with_visible(false);
     let config_template_builder = ConfigTemplateBuilder::new();
     let display_builder = DisplayBuilder::new().with_window_builder(Some(window_builder));
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -31,73 +31,96 @@ use std::env;
 use std::thread;
 use std::sync::{mpsc::Receiver, Mutex, Once, RwLock};
 
+// The code below down to `build_display` is a workaround due to a lack of a test initialization hook
+// This sort of design is recommended against for applications
+
+// Thread communication
+static mut EVENT_LOOP_PROXY: RwLock<Option<EventLoopProxy<()>>> = RwLock::new(None);
+static mut WINDOW_RECEIVER: Mutex<Option<Receiver<(Window, Config)>>> = Mutex::new(None);
+// Initialization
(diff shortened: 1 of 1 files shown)


justincredible:
I've updated the tests change so that it also works on X11 as well as Wayland and Windows. The issue with window visibility on Wayland is worse than before, but again does not affect the outcome of the tests. I don't remember why I tried this before, but on Windows I was able to achieve nearly invisible windows (they were still visible in the taskbar) using different options for the builders, and this would likely work for Wayland as well.

A couple problem areas are:
- This approach does require keeping in sync with the `raw_window_handle` and `winit` crates. However, whether `Window` or `RawWindowHandle` implement `Send` and `Sync` or not is likely pretty stable at this point.
- `Window`s created will stay alive until the test target completes, but I haven't encountered an issue with this yet. I suspect this can be worked around but not sure it's worth the effort.

The only systems I tested this on are Linux Mint (X11), Debian 12 (X11 and Wayland), and Window 10 (Win32). There are system-dependent test changes for [Mint](https://github.com/justincredible/glium/tree/unix-tests) and [Windows](https://github.com/justincredible/glium/tree/cargo-test-windows) that are not included. It's strange that the `todo` macro does not panic on Wayland.

est31 (approved):
Thanks!

justincredible:
Addendum: attempts to solve the tests issue with a more general approach

Everything below requires commit f140772, and assumes `tests::display::multiple_displays` is ignored.

### `cargo-xtask`
This works as an alternative to platform-specific scripts that call `cargo-test` on each individual test. The problem is reliably running every test and never more than one per process. One particularly troublesome issue is test function names that are prefixes of other tests in the same target.

The downside is likely needing to restructure the project into the typical workspace layout.

### `cargo-nextest`
This does work on Windows. With X11, this mostly works but targets `tests::{attrs, display, draw_parameters, texture_buffer, texture_creation, texture_sample, uniform_buffer, uniforms, vertex_buffer, vertex_source}` each contain at least one test that runs indefinitely, similar to [winit#2986](https://github.com/rust-windowing/winit/issues/2986).
Since Wayland mostly worked with the initial workaround, this might work there also but I did not test.

PR #2070, opened 2023-08-04
Doc test fixes

justincredible:
This mostly just updates the `Display`s to be generic in the doc tests. The other `src/macros.rs` doc test issues are addressed in separate commits. Only `src/vertex/buffer.rs` has a remaining test that fails when running

`cargo test --doc`

I'm not sure what the right approach for that doc test would be.


est31 (approved):
Thanks!

Issue #2069, opened 2023-08-01
Simplifying the sRGB and colorspace situation


jasoneveleth:
This issue is to follow up my previous issue #2059.

As it turns out, a lot of the implicit conversions are coming from OpenGL itself. See my [blog post](https://www.jasoneveleth.com/blog/2023/07/31/understanding-GLFRAMEBUFFERSRGB-in-glium). I summarized all the issues people have had with glium and colors. I made a flowchart (which is handwritten for now) which captures all the fixes. I also outlined two possible approaches to make using glium easier. I'll cover them here for convenience.

We can either hide the fact that OpenGL uses `GFS`, or we should be very transparent about what `GFS` means and how to interact with it.

## Transparent Approach

Here, `GFS` should be enabled or disabled when the context is initialized and never changed (there is already support for this in `ContextBuilder`). Using `clear_color()` to change the `GFS` value seems like the wrong way to go since it introduces unnecessary complexity. I can't think of a use case where being able to change `GFS` at runtime is useful if we're trying to be transparent (feel free to point one out in the comments).

In this approach with `GFS` enabled, we still require users to use `SrgbTexture`s and `Texture`s, and to differentiate what type of values are stored there. But, there is no need to specify whether a program outputs sRGB, since we are going to use OpenGL's underlying implicit conversions.

In this approach, if you choose to disable `GFS`: 
- you should do everything using linear RGB and no gamma correction will be applied (except from `SrgbTexture`, which will of course be converted to linear when sampled from)
- we assume the user understands what they are doing

### How to implement this approach

First, remove `clear_color_srgb()` and don't fiddle with `GFS` in `clear_color()`. Second, remove the code which enables `GFS` in `use_program()` at `5d50e70:src/program/program.rs:449`.

### Advantages of this approach

- familiarity: it respects the OpenGL model (using `GFS` as a global parameter) and obeys the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) 
- easy-of-use for beginners assuming they already understand OpenGL

### Disadvantages

- blending with an sRGB image won't work nicely with `GFS` disabled, so all data must be in linear color space, which is mildly inconvenient

## Hidden approach

The user should never have to worry about `GFS`. This is most similar to how things are done now. They do still need to worry about whether their textures and programs contain/output sRGB values.

### How to implement this approach

[...]

PR #2076, opened 2023-09-11
Fix sRGB confusion

justincredible:
Per the conversations from #2069 et cetera.

Part of the confusion is that `glutin` creates an sRGB default framebuffer, and there is apparently no way around this (for WGL and GLX) via the API in recent versions.

`clear_color` will likely remain confusing to new users, but hopefully the documentation update will prevent unnecessary issue tickets.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -784,12 +784,12 @@ pub trait Surface {
     fn clear(&mut self, rect: Option<&Rect>, color: Option<(f32, f32, f32, f32)>, color_srgb: bool,
              depth: Option<f32>, stencil: Option<i32>);
 
-    /// Clears the color attachment of the target.
+    /// Clears the color attachment of the target. The color is converted to sRGB when the target has sRGB format.
     fn clear_color(&mut self, red: f32, green: f32, blue: f32, alpha: f32) {
         self.clear(None, Some((red, green, blue, alpha)), false, None, None);
     }
 
-    /// Clears the color attachment of the target. The color is in sRGB format.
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -784,12 +784,12 @@ pub trait Surface {
...
              depth: Option<f32>, stencil: Option<i32>);
 
-    /// Clears the color attachment of the target.
+    /// Clears the color attachment of the target. Converts the color to sRGB when applicable.
     fn clear_color(&mut self, red: f32, green: f32, blue: f32, alpha: f32) {
         self.clear(None, Some((red, green, blue, alpha)), false, None, None);
     }
 
-    /// Clears the color attachment of the target. The color is in sRGB format.
+    /// Clears the color attachment of the target. The color is expected to match the target's color space.
```
est31, line 792: This is quite confusing to readers who wonder why the `_srgb` function does not expect data in srgb format. Could you extend the rustdoc to say what the `_srgb` stands for? (the srgb extension).

justincredible: Should I remove this line from the commit or change the the second sentences of the rustdocs to:
clear_color - "The color is converted to sRGB when the target has sRGB format."
clear_color_srgb -  "This function disables `GL_FRAMEBUFFER_SRGB`, so the color should be in sRGB format."

AFAICT `clear_color_srgb` will work with RGB input on an RGB target, but that's obviously not the intention.

est31: The first sentence is fine: "Clears the color attachment of the target.". I was talking more about the stuff after the first sentence.

est31 (approved):

PR #2079, opened 2023-10-04
fix: breaking typo

ben-sommer:
The example has a typo (`##` instead of `#`) when specifying the GLSL version number.


est31 (approved):
this was probably copied from a rustdoc comment because there you need to add two `#`es as one `#` indicates that it should be hidden.

PR #2093, opened 2024-01-01
SimpleWindowBuilder.build should be generic over EventLoop<T>

waynr:
This allows me to add my own events to the event loop via
`winit::event::Event`, prior to this change I was seeing errors that looked
like the following:

```
error[E0308]: mismatched types
   --> src/main.rs:62:16
    |
62  |         .build(&el);
    |          ----- ^^^ expected `&EventLoop<()>`, found `&EventLoop<MyEvent>`
    |          |
    |          arguments to this method are incorrect
    |
    = note: expected reference `&EventLoop<()>`
               found reference `&EventLoop<MyEvent>`
```

I haven't tried running any tests, just thought I would get feedback from
maintainers before doing any more.

### DIFF
--- a/src/backend/glutin/mod.rs
+++ b/src/backend/glutin/mod.rs
@@ -330,7 +330,7 @@ impl SimpleWindowBuilder {
 
     /// Create a new [`Window`](winit::window::Window) and [`Display`]
     /// with the specified parameters.
-    pub fn build(self, event_loop: &winit::event_loop::EventLoop<()>) -> (winit::window::Window, Display<glutin::surface::WindowSurface>) {
+    pub fn build<T>(self, event_loop: &winit::event_loop::EventLoop<T>) -> (winit::window::Window, Display<glutin::surface::WindowSurface>) {
         use glutin::prelude::*;
         use raw_window_handle::HasRawWindowHandle;
 


est31:
LGTM, I think it's good to have this change.

notgull:
I think a better solution would be to work a no-generic clonable owned display handle into `winit`, as proposed [here](https://github.com/rust-windowing/glutin/pull/1582#issuecomment-1858999854).

waynr:
> I think a better solution would be to work a no-generic clonable owned display handle into winit, as proposed https://github.com/rust-windowing/glutin/pull/1582#issuecomment-1858999854.

@notgull I'm still fairly new to graphics programming in general let alone glium/glutin/winit...but I don't understand how that solves the problem. Does obviate the need for `SimpleWindowBuilder` somehow such that I can use `winit::window::WindowBuilder` with its already generic `build` method?

Not disputing that your proposed solution is better, but is there anything particularly wrong with this approach that should prevent it being merged before the better-in-the-long-term solution is available? Is it possible someone is actually depending on `SimpleWindowBuilder` taking a non-generic `EventLoop<()>`?

notgull:
Ah my bad, misunderstood the problem. Yeah that wouldn't work

est31 (approved):

Issue #2059, opened 2023-06-15
neither clear_color_srgb nor clear_color_srgb displays the correct RGB color


jasoneveleth:
The discussion in #1793 seemed to make me believe that `clear_color_srgb` would fix my problems, but I didn't understand exactly how to call it. My code is here:
```rust
use glium::Surface;
use glutin::event::WindowEvent::CloseRequested;
use glutin::event::Event;

fn main() {
    let wb = glutin::window::WindowBuilder::new();
    let event_loop = glutin::event_loop::EventLoop::new();
    let cb = glutin::ContextBuilder::new();
    let window = cb.with_srgb(true).build_windowed(wb, &event_loop).expect("unable to create window");
    let display = glium::Display::from_gl_window(window).expect("unable to create display");

    event_loop.run(move |ev, _, control_flow| {
        match ev {
            Event::WindowEvent { event, .. } => match event {
                CloseRequested => {
                    *control_flow = glutin::event_loop::ControlFlow::Exit;
                    return;
                },
                _ => (),
            },
            Event::RedrawRequested(_) => {
                println!("INFO: redraw requested");
                let color = (0x28, 0x2c, 0x34);
                let (r, g, b) = ((color.0 as f32 / 255.0), (color.1 as f32 / 255.0), (color.2 as f32 / 255.0));
                let alpha = 1.0;

                let mut target = display.draw();
                target.clear_color_srgb(r, g, b, alpha);
                let _ = target.finish();
            }
            _ => (),
        }
    });
}
```

The color is lighter than it should be (just like in #1793). 

<details>
<summary>
I logged some of the context values using the diff in glium (click to exapand)
</summary>
DIFF:

```diff
diff --git a/src/context/mod.rs b/src/context/mod.rs
index e1e259c..072c1c5 100644
--- a/src/context/mod.rs
+++ b/src/context/mod.rs
@@ -2,6 +2,7 @@
 
 use crate::gl;
 
+use core::fmt;
 use std::collections::HashMap;
 use std::mem;
 use std::ptr;
@@ -43,6 +44,12 @@ mod extensions;
 mod state;
 mod uuid;
 
+impl<'a> fmt::Debug for CommandContext<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.state.fmt(f)
+    }
+}
+
 /// Stores the state and information required for glium to execute commands. Most public glium
 /// functions require passing a `Rc<Context>`.
 pub struct Context {
[...]

PR #2098, opened 2024-02-16
Add outputs_srgb example (Showing difference between sRGB and linear RGB)

faern:
I have been struggling a bit with rendering correct colors lately. As part of investigating this, I wanted to get a better understanding of what color spaces the shaders were expected to output. It was not intuitive to me how to change the sRGB configuration in glium, and there are not really any docs around it that I could find. All I found was #2069, talking about making it more intuitive. But that issue/PR did not add any docs or examples. It basically just changed the default value.

So I decided to create an example to show to myself how to do it and what difference it makes (gamma in this case). Figured I cold provide it upstream with these two intended outcomes:
* Get some feedback that improves my own own understanding of this topic
* Help others in the future

Ping @jasoneveleth and @justincredible, You might be interested in this since you worked on the previous sRGB issue/PR.

### DIFF
--- a/examples/srgb_shader.rs
+++ b/examples/srgb_shader.rs
@@ -1,0 +1,151 @@
+//! This example shows how to configure if a shader outputs linear RGB or sRGB
+//! color values.
+//!
+//! It draws two gradients from black to white. Both uses the exact
+//! same vertex position and color values. But one is rendered with outputs_srgb
+//! disabled and one with it enabled. This shows the visual difference of having
+//! gamma correction applied: bottom gradient, percieved as smooth linear transition
+//! from black to white. VS not having any gamma correction: top gradient, looks mostly
+//! white, with the blacks squashed into a small portion of the left side.
+//!
+//! Glium has sRGB enabled by default. This means that when you create a shader
+//! program with the `program!` macro, it will expect the shader to output sRGB
+//! color values.
+
+#[macro_use]
+extern crate glium;
+mod support;
+
+use glium::index::{NoIndices, PrimitiveType};
+use glium::{Display, Surface};
+use glutin::surface::WindowSurface;
+use support::{ApplicationContext, State};
+
+#[derive(Copy, Clone)]
+struct Vertex {
+    position: [f32; 2],
+    color: [f32; 3],
+}
+implement_vertex!(Vertex, position, color);
+
+fn create_program(display: &Display<WindowSurface>, outputs_srgb: bool) -> glium::Program {
+    let source = glium::program::ProgramCreationInput::SourceCode {
+        tessellation_control_shader: None,
+        tessellation_evaluation_shader: None,
+        geometry_shader: None,
+        outputs_srgb,
+        uses_point_size: false,
+
+        vertex_shader: "
+            #version 150
+
+            uniform mat4 matrix;
+            in vec2 position;
+            in vec3 color;
+            out vec3 vColor;
+
+            void main() {
+                gl_Position = vec4(position, 0.0, 1.0) * matrix;
+                vColor = color;
+            }
+        ",
+        fragment_shader: "
+            #version 150
+
+            in vec3 vColor;
+            out vec4 f_color;
+
+            void main() {
+                f_color = vec4(vColor, 1.0);
+            }
+        ",
+        transform_feedback_varyings: None,
+    };
+
+    glium::Program::new(display, source).unwrap()
+}
+
+struct Application {
+    pub vertex_buffer: glium::VertexBuffer<Vertex>,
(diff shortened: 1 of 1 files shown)


justincredible:
Jason's compilation comprises the context, but #2059 per se is a sufficient summary.

The `outputs_srgb` flag attests the shader program is responsible for outputting sRGB values (if necessary). Otherwise, `GL_FRAMEBUFFER_SRGB` is enabled and the shader output is automatically converted when writing an sRGB target.

This is certainly easier to explain with a visualization, but the comments at the top of the file may be a bit confused. `glium` merely toggles the OpenGL colorspace correction and the result is due to `glutin`'s provided default framebuffer.

faern:
> `glium` merely toggles the OpenGL ...

But I don't know that part of OpenGL and was hoping I could learn it through the way I'm consuming it (`glium`). Especially since it has some built in magic in between me and OpenGL that is not 100% obvious to me.

I'm open to suggestions how to better word the description for the example! Or other ways of implementing it that makes it easier to understand.

> The outputs_srgb flag ... Otherwise, GL_FRAMEBUFFER_SRGB is enabled

This was not really obvious to me. And the documentation for `SpirvProgram::outputs_srgb` states: "See SourceCode::outputs_srgb.". But `SourceCode` has no member of that name. Updating this documentation to be more verbose would be amazing. I would do it myself if I knew what to write, but I think you have a much better understanding of that.

The best would of course be to be able to visualize the pipelines how color values travel from wherever they start to the actual physical monitor. Where gamma correction is added/reversed and what more steps there are. Personally I don't care too much about textures. I have color values in XYZ-format that I get from various colorimetry tools. And I want to write software that renders these values as correct as possible (I'm aware that anything outside the sRGB gamut won't be possible to display of course).

faern:
@justincredible Do you have the source code for your four examples available? When you say you enable and disable color correction, what are you referring to exactly?

justincredible:
There is a git-diff below the image, but I've also provided the altered source [here](https://github.com/justincredible/glium/tree/pr2098). There is another level to explore with how the background is cleared for both types of textures.

The `SpirvProgram` docs are referring to these [docs](https://docs.rs/glium/latest/glium/program/enum.ProgramCreationInput.html#variant.SourceCode). I have not used SPIR-V myself.
The sRGB behaviour is discussed in the [`texture`](https://docs.rs/glium/latest/glium/texture/index.html#about-srgb) docs.
The other point of control (besides shader programs) is the [`Surface::clear`](https://docs.rs/glium/latest/glium/trait.Surface.html#tymethod.clear) function with the `color_srgb` flag. Most users enable the [GL_FRAMEBUFFER_SRGB](https://www.khronos.org/opengl/wiki/Framebuffer#Colorspace) color correction by calling `clear_color`.

I'm relatively new to this library and the practicalities of sRGB, so I may be ignorant of some wider context.

faern:
> The SpirvProgram [docs](https://docs.rs/glium/latest/glium/program/enum.ProgramCreationInput.html#variant.SourceCode) are referring to these docs. I have not used SPIR-V myself.

Aaah. I did not find those. If you search for `outputs_srgb` in the rustdoc they turn up in a very unhelpful order. The first one is the `SpirvProgram::outputs_srgb` build method, which has *no valuable docs*. Second hit is the `SpirvProgram::outputs_srgb` member, which up until #2099 was not clickable. My separate documentation PR aims to hopefully slightly improve on this :)

> The sRGB behaviour is discussed in the [texture](https://docs.rs/glium/latest/glium/texture/index.html#about-srgb) docs.

Those docs must surely be outdated, no? Because they say:

> By default, glium enables the GL_FRAMEBUFFER_SRGB trigger, which expects the output of your fragment shader to be in linear RGB 

While the `outputs_srgb` flag says the flag is enabled by default, and that `GL_FRAMEBUFFER_SRGB` is enabled only when `outputs_srgb` is disabled. So they contradict each other right now.

faern:
For my own understanding, what's the correct/simplest way to just render linear RGB values directly to the window framebuffer? I don't have any textures. I just have linear RGB data that I want to render as simple but correct as possible. I can encode them to srgb also, but if I can avoid that step and have OpenGL do whatever color correction is needed that's just simpler.

My understanding is that I should then do `clear_color` for clearing the framebuffer and set `outputs_srgb: false` in my shader program?

justincredible:
`clear_color` and `outputs_srgb: false` each enable color correction for their operation, but this only affects sRGB targets.

`clear_color_srgb` and `outputs_srgb: true` are what you want for the window framebuffer.

faern:
Given that my shader writes srgb values, no? What if my shader writes linear rgb to the display framebuffer?

PR #2101, opened 2024-02-21
Publicly export `Capabilities` and `ExtensionsList` struct

faern:
It was already exposed via the `CapabilitiesSource::get_capabalities` method anyway, but the struct itself was not public. So a library user could not refer to the type, neither could they click their way through to the struct documentation.

See https://docs.rs/glium/latest/glium/trait.CapabilitiesSource.html#tymethod.get_capabilities


faern:
Oh yeah. Not that I needed it now. But I saw `ExtensionsList` were in the same state. So I publicly re-exported that also!

est31 (approved):

Issue #2106, opened 2024-03-24
Unable to draw texture in `gtk4-rs` application


faern:
I have been able to render OpenGL nicely in `GLArea`s in a `gtk4-rs` application in general. But using textures simply does not work for me. I [reported this to gtk4-rs](https://github.com/gtk-rs/gtk4-rs/issues/1647), but I figured the issue might be with `glium`, so I report it here also.

I get the image texture examples from `glium` to run just fine! But when I port that exact same glium code over to render in a GTK4 application, the textures simply don't show up. It's as if the `texture(...)` function in the fragment shader just returns transparent. The GL context has `DebugCallbackBehavior::DebugMessageOnError` and no errors or warnings are printed.

I have created a single-file example of my problem (plus corresponding `Cargo.toml`). This example is initially based off of the [glium `image` example](https://github.com/glium/glium/blob/master/examples/image.rs) merged with the [gtk4-rs OpenGL example](https://github.com/gtk-rs/gtk4-rs/tree/master/examples/glium_gl_area).

<details>
<summary>Cargo.toml</summary>

```toml
[package]
name = "gtk_glium_texture2"
edition = "2021"

[dependencies]
gtk4 = "0.8.0"
glium = "0.34.0"
gl_loader = "0.1.2"
image = "0.25"
```
</details>

<details>
<summary>main.rs</summary>

```rust
use core::ffi::c_void;
use glium::backend::{Backend, Context, Facade};
use glium::debug::DebugCallbackBehavior;
use glium::index::{NoIndices, PrimitiveType};
use glium::texture::{RawImage2d, Texture2d};
use glium::{
    implement_vertex, program, uniform, Frame, IncompatibleOpenGl, Program, Surface,
    SwapBuffersError, VertexBuffer,
};
use gtk4::prelude::*;
use gtk4::{Application, ApplicationWindow, GLArea};
use std::io::Cursor;
use std::rc::Rc;
use std::time::Duration;

fn main() {
    let application = Application::builder()
        .application_id("com.example.Gtk4GliumTexture")
        .build();

    application.connect_activate(|app| {
        let window = ApplicationWindow::builder()
            .application(app)
            .title("Texture")
            .default_width(600)
            .default_height(400)
            .build();

        let glarea = GLArea::builder().vexpand(true).build();

        window.set_child(Some(&glarea));
        window.show();

        let facade = GtkFacade::from_glarea(&glarea).unwrap();

        let opengl_texture = load_texture(&facade);
[...]

PR #2110, opened 2024-05-15
Add exec_with_context to make repair possible when glium's invariants have been broken

awused:
This can be used when another component has already modified the OpenGL context out from underneath glium. Even more than exec_in_context, the application needs to take responsibility to maintain glium's invariants, but it can be used to restore them when they've already been broken.

This can be used to address https://github.com/glium/glium/issues/2106, barring a better solution. It's very unsafe but there's not too much an application could do to break glium more with this than with `exec_in_context` if they wanted to try.


src/context/mod.rs
```diff
@@ -542,6 +542,22 @@ impl Context {
...
+    /// glium's internal state. Useful if another component needs to directly manipulate
+    /// OpenGL state or if another component already has.
+    ///
+    /// **`action` takes full responsibility for keeping OpenGL state in sync with the
+    /// context's state.**
+    #[inline]
+    pub unsafe fn exec_with_context<'a, T, F>(&self, action: F) -> T
+    where
+        T: Send + 'static,
+        F: FnOnce(&CommandContext<'_>) -> T + 'a,
```
awused, line 555: I opted not to make this a &mut as I couldn't see an immediate need for it, but in theory it wouldn't make it that much worse.

awused: Actually making a more general example for the bug report was a use for &mut.

est31:
I wonder if for https://github.com/glium/glium/issues/2106 one could make two separate OpenGL contexts in the same application.

awused:
My bad on the spate of force commits, I noticed I hadn't actually pushed the update with &mut ctxt, then realized I'd formatted the entire file.

awused:
> I wonder if for #2106 one could make two separate OpenGL contexts in the same application.

Maybe for this specific case; I'm honestly not familiar enough with GTK and opengl to say. But I think this PR is useful in its own right. If you ever need to step outside of glium and you want to restore the opengl context to what glium thinks it is, or vice versa, it's helpful to know what glium thinks the state is. With exec_in_context, if I need to bind a texture to do something unusual, I don't know what texture to bind to that unit after I'm done without explicitly querying the gpu myself.

remcokranenburg:
This would indeed help a lot with interop!

One thing that comes to mind here is that `exec_with_context` is almost the same as `exec_in_context`. Since glium is still 0.x, maybe it's better to just modify the original function instead of adding more API.

awused:
Rather than making a breaking change, I think it'd make sense to deprecate `in_context` instead of changing it, but I'm not going to do that myself, and `in_context` is still sometimes enough for things like uploading textures. But they are pretty redundant and I do think the `with_context` method makes more sense than `in_context` if glium only wanted one, because it's hard to uphold the requirements of `in_context` without knowing what glium thinks the context is.

PR #2115, opened 2024-07-06
Re-export winit

### LINKED ISSUE #2086: Re-export winit
There are a lot of public exposed APIs that accept and return winit types (ie):

https://github.com/glium/glium/blob/1676f0de6dddfa294f834613372f5096a429acd8/src/backend/glutin/mod.rs#L333

Also, most of the tutorials make heavy use of the winit APIs.

Currently this requires users to figure out which version of winit is currently being used, and make sure their winit dependency stays in sync with the glium dependency, otherwise it will be a breaking change if your API expects a 0.29 EventLoop when the user is still using a 0.28 EventLoop.

It seems like you are aware of this, and are bumping semver minor to signify pre-1.0.0 breaking changes whenever you bump the winit dependency to a breaking change.

Since that is already a requirement on your library, you might as well re-export winit explicitly and change the tutorials to recommend using the re-export.

junderw:
Closes #2086 

I ran all the checks in the CI yaml, but actually running clippy and `cargo test` (actually running the tests) gave me tons of errors in places I didn't touch.

### DIFF
--- a/book/tuto-01-getting-started.md
+++ b/book/tuto-01-getting-started.md
@@ -38,9 +38,11 @@ Initializing a simple OpenGL window with the default winit/glutin backend can be
 
 This will open a new window, register it with the given event_loop and create a (glutin) OpenGL context and glium Display while finally returning both the window and display to you.
 
+The following examples require the `simple_window_builder` feature, and provide a re-export of winit to prevent version mismatches.
+
 ```rust
 fn main() {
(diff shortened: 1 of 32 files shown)


junderw:
beta and nightly failures are unrelated...

I wonder what caused them...

est31:
> I wonder what caused them...

I have fixed the warnings in #2116.

junderw:
Rebased on master

est31 (approved):

est31:
Thanks for taking the time to make a PR!

PR #2120, opened 2024-08-17
Support border color for texture sampler

shouya:
Allow specifying for GL_TEXTURE_BORDER_COLOR, which is often used in shadow mapping.

### DIFF
--- a/src/uniforms/sampler.rs
+++ b/src/uniforms/sampler.rs
@@ -159,6 +159,13 @@ impl From<BorderColor> for [f32; 4] {
     }
 }
 
+impl BorderColor {
+    #[inline]
+    pub(crate) fn as_f32_array(&self) -> &[f32; 4] {
+        unsafe { std::mem::transmute(&self.0) }
+    }
+}
+
 /// A sampler.
 #[derive(Debug, Hash, PartialEq, Eq)]
 pub struct Sampler<'t, T>(pub &'t T, pub SamplerBehavior);
--- a/src/uniforms/sampler.rs
+++ b/src/uniforms/sampler.rs
@@ -140,6 +140,25 @@ impl ToGlEnum for DepthTextureComparison {
     }
 }
 
+
+/// The border color to use for `SamplerWrapFunction::BorderClamp`.
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


src/uniforms/sampler.rs
```diff
@@ -140,6 +140,36 @@ impl ToGlEnum for DepthTextureComparison {
...
+/// This type encapsulates `[f32; 4]` internally as `[u32; 4]` so that
+/// it implements Hash and Eq. The conversions to and from `[f32; 4]`
+/// are provided by the `From` trait.
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub struct BorderColor([u32; 4]);
+
+impl From<[f32; 4]> for BorderColor {
+    #[inline]
+    fn from(value: [f32; 4]) -> BorderColor {
+        BorderColor(unsafe { std::mem::transmute(value) })
```
est31, line 155: there is no need to use unsafe here: you can use array.map with `f32::to_bits`. in the end it will most likely compile to the same code.

src/uniforms/sampler.rs
```diff
@@ -140,6 +140,36 @@ impl ToGlEnum for DepthTextureComparison {
...
+    }
+}
+
+impl From<BorderColor> for [f32; 4] {
+    #[inline]
+    fn from(value: BorderColor) -> [f32; 4] {
+        unsafe { std::mem::transmute(value.0) }
```
est31, line 162: same here

src/uniforms/sampler.rs
```diff
@@ -140,6 +140,36 @@ impl ToGlEnum for DepthTextureComparison {
...
+    }
+}
+
+impl BorderColor {
+    #[inline]
+    pub(crate) fn as_f32_array(&self) -> &[f32; 4] {
+        unsafe { std::mem::transmute(&self.0) }
```
est31, line 169: same here

est31 (approved):

src/uniforms/sampler.rs
shouya: it seems that this one is not applicable to the array.map + f32::from_bits trick. any suggestions?

est31: ah yeah, good point. let's leave it as is.

PR #2121, opened 2024-09-07
Fix examples and tests for Windows

justincredible:
Examples
-
The underlying API used by [`request_redraw()`](https://docs.rs/winit/0.30.5/winit/window/struct.Window.html#method.request_redraw) does not produce any `WM_PAINT` messages for invisible windows, so these windows do not receive the `RedrawRequested` event. This leaves the two examples `info` and `gpgpu` running indefinitely with no clean way to exit. This change requires any future, invisible, run once examples to also complete by the time `State::new()` returns like the current two do.

Tests
-
With `raw-window-handle-0.6` Windows handles now implement `Send + Sync`. The tests workaround is erroring when passing a `RefWindow`:
```
thread 'basic' panicked at tests\support\mod.rs:83:73:
called `Result::unwrap()` on an `Err` value: Unavailable
``` 

OHOS was added in v0.6.2 and can cause build errors if an older version is cached.


est31 (approved):

est31:
Thanks for these great PRs, @justincredible !

PR #2126, opened 2024-09-27
Add ActiveEventLoop support for SimpleWindowBuilder

OLEGSHA:
winit has [deprecated `EventLoop::create_window`](https://docs.rs/winit/0.30.0/winit/event_loop/struct.EventLoop.html#method.create_window) and [deprecated `EventLoop::run`](https://docs.rs/winit/0.30.0/winit/event_loop/struct.EventLoop.html#method.run) since 0.30.0. This means that an `&EventLoop` is no longer accessible without warnings in code that should call `SimpleWindowBuilder::build`.

The simple solution of replacing `&EventLoop` with `&ActiveEventLoop` is out of question because that would be a breaking change. Adding a new `build` method with a new name is extremely inelegant.

Luckily, `glutin_winit` already supports using both. This PR deals with the situation by introducing a new trait, `GliumEventLoop`, that is implemented for both `EventLoop` and `ActiveEventLoop`, and changing `::build()` to accept an `&impl GliumEventLoop`. This is the same approach that `glutin_winit` uses. Unfortunately, the trait from `glutin_winit`, [`GlutinEventLoop`](https://docs.rs/glutin-winit/latest/src/glutin_winit/event_loop.rs.html#11), is not public, so it cannot be reused directly and must be effectively duplicated.

### DIFF
--- a/src/backend/glutin/mod.rs
+++ b/src/backend/glutin/mod.rs
@@ -31,6 +31,12 @@ use std::ops::Deref;
 use std::os::raw::c_void;
 use std::rc::Rc;
 
+#[cfg(feature = "simple_window_builder")]
+pub use self::simple_window_builder::SimpleWindowBuilder;
+
+#[cfg(feature = "simple_window_builder")]
+pub mod simple_window_builder;
+
 /// Wraps a glutin context together with the corresponding Surface.
 /// This is necessary so that we can swap buffers and determine the framebuffer size within glium.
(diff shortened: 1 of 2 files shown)


est31 (approved):

OLEGSHA:
Huh. Am I correct that these failures are not caused by my changes directly? Should I fix them?

est31:
@OLEGSHA if you want you could make a separate PR but they are unrelated indeed.

PR #2127, opened 2024-09-28
Fix `static_mut_refs` warnings

OLEGSHA:
Fix for nightly check warnings discovered in #2126.

Turns out, most of `static mut`'s are not needed in the first place, and the one that is necessary is a dirty hack regardless.

### DIFF
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -173,20 +173,17 @@ pub fn build_display() -> Display<WindowSurface> {
     unsafe { initialize_event_loop(); }
 
     // Tell event loop to create a window and config for creating a display
-    unsafe {
-        EVENT_LOOP_PROXY
-            .read().unwrap()
-            .as_ref().unwrap()
-            .send_event(()).unwrap();
-    }
+    EVENT_LOOP_PROXY
+        .read().unwrap()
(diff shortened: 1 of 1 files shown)


est31 (changes requested):
Could you make `WINDOWS` use a `Mutex`? This is the more future-proof way of addressing the lint.

OLEGSHA:
I don't know how to make it work. If I wrap `WINDOWS` in a `Mutex`, then the `window` borrow on line 141 is no longer 'static because it originates from a locked mutex:
```rs
let mut mutex_lock = WINDOWS.lock().unwrap();
let window = &mutex_lock.as_mut().unwrap()[&key];
sender.send((window.into(), gl_config)).unwrap();
// `mutex_lock` dropped here while still borrowed
```

OLEGSHA:
If I may express my opinion here, I think the use of a global variable here is the core issue. Ideally, the tests should be refactored to eliminate the need for it in the first place. I'm not yet ready to commit myself to this refactor, and I understand that no one else is available. I would then suggest that the awkward and annoying `#[allow]` band-aid is kept around so that this solution does not look prettier than it actually is.

In personal experience, if you want something to be temporary, make it ugly, so it's impossible to forget.

PR #2128, opened 2024-10-05
Leak Window object via Box::leak instead of HashMap

est31:
The prior code was UB: it's possible that the hash map can move around the objects while it grows. We were leaking previously as well, so this is not much of a regression.

cc prior PR #2127


OLEGSHA:
As far sa I can tell, this prevents the destructor of `Window` from ever running. Is this safe on all platforms?

est31:
static doesn't run destructors either, so it's not a regression. Whether destructors are required or not at all, I don't know. But I'd put that onto a future refactor.

Issue #2118, opened 2024-08-03
How to enable vsync?


scurest:
How do you enable vsync/set_swap_interval in current glium?

I am trying to upgrade a project to the latest version. As of at late as v0.32, you would just add `.with_vsync(true)` to [the `glutin::ContextBuilder`](https://github.com/glium/glium/blob/v0.32.0/examples/tutorial-01.rs#L7C1-L7C44), but I don't see how to do it with the new `SimpleWindowBuilder`.

scurest:
For reference, [this patch](https://github.com/scurest/glium/commit/f59f3ed58044a088bdc8668ac927387999a51e44) seems to work for me.

est31:
PRs welcome!

nico-abram:
In the meantime, I was able to do it by copying the code from `SimpleWindowBuilder::build` and adding this line:

```rs

surface.set_swap_interval(&current_context, glutin::surface::SwapInterval::DontWait).unwrap();

```

For glium 0.34 I had to add these dependencies:

```toml
glutin = "0.31.3"
raw-window-handle = "0.5.2"
glutin-winit = "0.4.2"
```

The resulting code looked like this:
(I actually wanted to *disable* vsync, which seemed to be enabled by default for me. Just change the value passed to set_swap_interval)


```rs

  let (window, display) = {
    use glutin::{display::GetGlDisplay, prelude::*};
    use raw_window_handle::HasRawWindowHandle;

    // First we start by opening a new Window
    let display_builder = glutin_winit::DisplayBuilder::new().with_window_builder(Some(builder));
    let config_template_builder = glutin::config::ConfigTemplateBuilder::new();
    let (window, gl_config) = display_builder
      .build(&event_loop, config_template_builder, |mut configs| {
        // Just use the first configuration since we don't have any special preferences here
        configs.next().unwrap()
      })
      .unwrap();
    let window = window.unwrap();

    // Now we get the window size to use as the initial size of the Surface
    let (width, height): (u32, u32) = window.inner_size().into();
    let attrs = glutin::surface::SurfaceAttributesBuilder::<glutin::surface::WindowSurface>::new()
      .build(
        window.raw_window_handle(),
        std::num::NonZeroU32::new(width).unwrap(),
        std::num::NonZeroU32::new(height).unwrap(),
      );

    // Finally we can create a Surface, use it to make a PossiblyCurrentContext and create the glium Display
[...]

PR #2131, opened 2024-10-12
Add vsync configuration to SimpleWindowBuilder

nico-abram:
Fixes https://github.com/glium/glium/issues/2118


est31 (approved):

OLEGSHA:
It seems a little limiting to me that the only option is set `SwapInterval` to 1. Why not have `with_swap_interval(mut self, u32)` (0 = `DontWait`) and a `with_vsync(mut self) -> Self { self.with_swap_interval(1) }`?

est31:
@OLEGSHA that would be a good followup I think

PR #2135, opened 2024-10-19
Clean up test workaround

justincredible:
Thanks for pointing out and fixing the UB in #2128. This workaround can be made safe (besides the glutin functions), and some API updates improve it further:
- Using the new winit API doesn't change much, but choosing a useful event type cleans up the thread communication. Also the event loop proxy setup was unnecessarily complex.
-  Since `glutin-0.31.3` the not current context and surface can both be sent to another thread. This avoids some cross-platform workarounds with sending the config.
- [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) is only recently stable, but makes the event loop initialization much more readable.

### DIFF
--- a/tests/support/mod.rs
+++ b/tests/support/mod.rs
@@ -6,27 +6,29 @@ Test supports module.
 #![allow(dead_code)]
 
 use glium::Display;
-use glium::backend::Facade;
+use glium::backend::{Context, Facade};
 use glium::index::PrimitiveType;
 
-use glutin::config::{Config, ConfigTemplateBuilder};
-use glutin::context::ContextAttributesBuilder;
+use glutin::config::ConfigTemplateBuilder;
+use glutin::context::{ContextAttributesBuilder, NotCurrentContext};
 use glutin::display::GetGlDisplay;
 use glutin::prelude::*;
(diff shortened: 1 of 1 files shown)


tests/support/mod.rs
```diff
@@ -6,197 +6,142 @@ Test supports module.
...
-                            SEND_PROXY.call_once(|| {
-                                ots.send(proxy.clone()).unwrap();
-                            });
-                        }
-                    }
-                })
-                .unwrap();
-            })
-            .unwrap();
+pub struct WindowDisplay {
```
justincredible, line 109: The window can be kept alive in the app, but sending it to the test thread improves the DX on Wayland as non-visible windows are still [unsupported](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_visible).

Using a smart pointer instead of a record or tuple avoids updating the `build_display` call sites and is otherwise unnecessary.

est31:
> [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) is only recently stable

So we'd have an MSRV of 1.80.0 for the tests? Eventually we can have, but I'd like to have at least some additional versions since then.

Could you keep out LazyLock changes from this PR?

est31 (approved):

Issue #1918, opened 2021-03-10
Trying to initialize dynamically sized SSBO


bjadamson:
Hi, for the past few days I've been trying to figure out how to pass a `[T]` through to my shaders, as the last parameter of the SSBO.

ie:

```glsl
// shader code
layout (std140) buffer BlockName
{
  mat4 manyManyMatrices[];
};
```

The smallest example I could find is [here](https://github.com/glium/glium/blob/0a17cd22d317abbf62458336986390a57649ff94/src/buffer/mod.rs#L28-L44) that shows how to do it. It however doesn't show how to pass it to the `uniform!` macro, which is what I'm trying to do. From what I understand that is the way to pass all data to my shader as an SSBO.

I've looked through all the issues within this repo, and that I can find through google on this. In the past it seemed enough to call `implement_buffer_content` and `implement_uniform_block` on the data. ie:
* [SO](https://stackoverflow.com/a/33003172/562174)
* [issue](https://github.com/glium/glium/issues/1718)

However when I try and compile the same code today, I get an error from `implement_uniform_block!` about values being unsized:

```rust
struct Data {
    values: [u8],
}
implement_buffer_content!(Data);
implement_uniform_block!(Data, values);

...
let mut buffer: glium::uniforms::UniformBuffer<Data> =
              glium::uniforms::UniformBuffer::empty_unsized(&display, 4 * 128).unwrap();
frame.draw(&vertices_buffer, &indices_buffer, &shader_program,
                   &uniform! { uniform_data: &buffer },
                   &Default::default()).unwrap();
```
> error[E0599]: the function or associated item `uninit` exists for union `MaybeUninit<Data>`, but its trait bounds were not satisfied
> 50 |         struct Data {
>    |         ----------- doesn't satisfy `Data: Sized`
> ...
> 57 |         implement_uniform_block!(Data, power, values);
>    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item cannot be called on `MaybeUninit<Data>` due to unsatisfied trait bounds
>    |
>    = note: the following trait bounds were not satisfied:
>            `Data: Sized`

My best guess to work around this would be to use a Box, but it UniformBlock isn't implemented for Box<[u8]> (or more importantly Box<[T]>).

```rust
struct Data {
    values: Box<[u8]>,
}
implement_buffer_content!(Data);
implement_uniform_block!(Data, values);

[...]

PR #2143, opened 2025-05-09
Use the std::mem::offset_of!() macro to calculate field offsets

bschwind:
Rust 1.86 [added more null checks](https://github.com/rust-lang/rust/pull/134424) when debug assertions are enabled which was causing this code in the `implement_uniform_block!()` macro to throw a null pointer dereference error.

Here's a small [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5acca75733b1496472979fa9ba3c0b12) which demonstrates the problem. Run it in debug mode, or run miri on it.

As far as I can tell, the original code wanted two things:

* The offset of the field in question, which the standard library now has an answer for (and glium was also already pulling in a crate which did this, not sure why it wasn't used)
* A valid reference to `$struct_name.$field_name`, so it can pass it to `matches_from_ty`, cast it to a `UniformBlock`, and recursively call `matches()` on it.

I verified that this fixed the null pointer dereference in my project's code which was using this macro, but I might be missing something.

One difference with my code is it constructs an entire zeroed instance of the struct, instead of just a zeroed _pointer_ to the struct. There might be a performance penalty to that which the original authors were trying to avoid, I'm not sure.

Related to https://github.com/glium/glium/issues/1918#issuecomment-1433533902

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -431,7 +431,7 @@ macro_rules! implement_uniform_block {
                             }
                         }
 
-                        fn matches_from_ty<T: $crate::uniforms::UniformBlock + ?Sized>(_: &T,
+                        fn matches_from_ty<T: $crate::uniforms::UniformBlock + ?Sized>(_: Option<&T>,
                             layout: &$crate::program::BlockLayout, base_offset: usize)
                             -> ::std::result::Result<(), $crate::uniforms::LayoutMismatchError>
                         {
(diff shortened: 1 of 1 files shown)


est31 (approved):

bschwind:
@est31 one thing I'm not sure about with my change is whether or not it breaks some use cases around Dynamically Sized Types (DSTs). I need to research a bit more because that might be a use case which gets broken with this change.

est31:
From my understanding, `offset_of` has DST support, in the sense that if you use a field that is before a DST field you can get its offset. If there is any breakage, I think it's okay to push to our users. 

Beyond that, there is definitely still unsoundness via the `matches_from_ty` function, as it accesses zeroed memory. Better would be to do sth like:

```Rust
    struct V { a: u8, b: u8, c: u8, }
    fn matches_from_ty<T>(_: Option<&T>) {}

    let opt = None::<&V>.map(|v| &v.a);
    matches_from_ty(opt);
```

bschwind:
Ohh the `Option` trick there is nice, let me see if I can work that in.

PR #2144, opened 2025-06-27
Add ability to set swap interval (Vsync)

djgaven588:
``set_swap_interval`` is available in the underlying glutin implementation, exposing it here allows an application to set vsync without recreating the ``Display<WindowSurface>``.

Also has a small edit to the ``swap_buffers`` description to correctly match it's underlying description.

### DIFF
--- a/examples/manual-creation.rs
+++ b/examples/manual-creation.rs
@@ -17,28 +17,26 @@ There are three concepts in play:
 
 */
 
-use winit::event_loop::EventLoop;
 use glium::Surface;
-use glutin::surface::WindowSurface;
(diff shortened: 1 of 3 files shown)


est31 (approved):

djgaven588:
Hey @est31 noticed I missed something, seems the examples aren't always compiled when testing.

Forgot a single unwrap:
https://github.com/djgaven588/glium/commit/db2dd2b06b6a474159f91142e52c1c91666a6a95

Would you like me to create a PR? (Don't want to make excessive ones)

PR #2147, opened 2025-07-09
Run tests in CI

justincredible:
Add ~30s for windows and 20-30s for ubuntu in CI jobs. The change to the book job is itself useful even if running the tests is not desirable.

The following tests are false positives on free-tier runners as they lack the capabilities to execute the test:
| Target | Test | Only on Windows |
| - | - | - |
| draw_parameters | provoking_vertex_first | |
| | provoking_vertex_last | |
| shaders | complex_layout | |
| | get_program_binary | Y |
| | program_binary_reload | Y |
| | program_binary_working | Y |
| ssbo | basic | |
| | custom_bind_point | |
| texture_creation | bindless_texture_residency_context_rebuild | |
| texture_sample | bindless_texture | |
| uniform_buffer | block | |
| | block_wrong_type | |
| | persistent_block_race_condition | |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,33 @@ jobs:
         RUSTFLAGS: -D warnings
       run: |
          cargo test --all --all-targets --no-run
+    - name: Install Xvfb
+      if: (matrix.os == 'ubuntu-latest')
+      run: |
+        sudo apt-get install libxkbcommon-x11-0
+        sudo apt-get install xvfb
+    - name: Run the tests (X11)
+      if: (matrix.os == 'ubuntu-latest')
+      env:
+        RUST_BACKTRACE: 1
+        RUSTFLAGS: -D warnings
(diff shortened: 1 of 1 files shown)


est31:
Will this work? I've thought glutin forbids context creation multiple times in the process. it is only allowed to be done once, which means that we can't run tests in parallel. IIRC that's why we disabled them.

justincredible:
Yes, it works the same as running the tests locally (except for being headless). However, comparing a [typical run](https://github.com/justincredible/glium/actions/runs/16129977336/job/45515355470) to [another one](https://github.com/justincredible/glium/actions/runs/16182572941/job/45682097394) with the additional flags ` -- --test-threads 1`, there is only a small timing difference but it is probably still much faster than running each test in its own process.

> glutin forbids context creation multiple times in the process

The only thing I'm aware of that needs to be unique per process is the event loop. The `multiple_windows` example creates a new context and surface for each new window.

PR #2152, opened 2025-08-29
Add support for RGB565

bschwind:
This is quite a common format for embedded displays and such, I'm not sure why it was originally left as a TODO in the code (maybe some gotchas I'm not aware of?)

In any case, I've tested this and it _does_ produce valid RGB565 data. The part I'm somewhat unsure of are the _exact_ versions of GL and GLES which added support for this format. Someone please double check me on that.

This has been tested on MacOS and Linux.

### DIFF
--- a/src/image_format.rs
+++ b/src/image_format.rs
@@ -205,7 +205,6 @@ impl ClientFormat {
 /// Some formats are marked as "guaranteed to be supported". What this means is that you are
 /// certain that the backend will use exactly these formats. If you try to use a format that
 /// is not supported by the backend, it will automatically fall back to a larger format.
-// TODO: missing RGB565
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub enum UncompressedFloatFormat {
     ///
@@ -246,6 +245,8 @@ pub enum UncompressedFloatFormat {
     U4U4U4,
     ///
     U5U5U5,
+    ///
+    U5U6U5,
     ///
     ///
     /// Guaranteed to be supported for textures.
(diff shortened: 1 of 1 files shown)


src/image_format.rs
```diff
@@ -435,6 +436,9 @@ impl UncompressedFloatFormat {
             UncompressedFloatFormat::U5U5U5 => {
                 version >= &Version(Api::Gl, 1, 1) || version >= &Version(Api::GlEs, 3, 0)
             },
+            UncompressedFloatFormat::U5U6U5 => {
+                version >= &Version(Api::Gl, 4, 0) || version >= &Version(Api::GlEs, 2, 0)
```
mbernat, line 440: tl;dr: looks like this should be 4.2.

```suggestion
                version >= &Version(Api::Gl, 4, 2) || version >= &Version(Api::GlEs, 2, 0)
```

This document mentions OpenGL 4.1, FWIW. https://www.khronos.org/opengl/wiki/Image_Format#Required_formats

> GL_RGB565 OpenGL 4.1 or [ARB_ES2_compatibility](http://www.opengl.org/registry/specs/ARB/ES2_compatibility.txt)

But checking the specs, 565 is not mentioned in 4.1, only in 4.2. In fact, it looks like it was not even present in the earliest version of the spec but only added to 4.2 later. https://registry.khronos.org/OpenGL/specs/gl/glspec42.core.pdf

> Add RGB565 format to table 3.12 (Bug 8530)

This ARB_ES2 discussion also appears to confirm that the earliest version is indeed 4.2, although the phrasing is rather confusing. https://registry.khronos.org/OpenGL/extensions/ARB/ARB_ES2_compatibility.txt

>     (12) Should RGB565 be added to the extension?
>    RESOLVED: Yes. The initial version of this extension did not include it,
    as did OpenGL 4.1, which included functionality from this extension.
    This was an oversight. RGB565 was added to the required format list in
    the initial release of the OpenGL 4.2 spec, but it was not added to the
    table of sized internal color formats. However, RGB565
    compatibility is important to ES2 compatibility.
>    The ARB discussed this situation in February 2012 and agreed to revise
    the extension spec and OpenGL 4.2 spec to fully support RGB565. While
    some drivers may not support RGB565, we believe they will add support
    quickly.

mbernat: Oh, but macos is on 4.1 and it works there.

bschwind: > Oh, but macos is on 4.1 and it works there.

Yep, that's what had me hesitate and place it at 4.0 for now. It would be nice to not exclude MacOS since it technically works.

mbernat: I think it would be okay to use 4.1 for macos (assuming it works) and 4.2 elsewhere.

The min version can be reduced later, if people find 4.1 works everywhere despite the spec but it would not be correct to claim 4.0 is supported and then crash on those old runtimes.

